Issue #4, opened 2017-12-05
Make all commands apply schema (if not exist)


bzz:
Right now, only tests do not rely on having existing schema in DB.

Each command should be changed (`hash`, `query`, `report`) to do the same and create schema, instead of failing if it does not exist.

PR #5, opened 2017-12-07
Hotfix: make CI pass

bzz:
This includes
 - [x] use ScyllaDB in Docker container instead of EmbeddedCassandra in test
 - [x] run integration tests with actual Apache Spark standalone installation

Let's merge ASAP as a hotfix, as soon as CI is green.


bzz:
Current CI failure is due to unable to connect to ScyllaDB

```
[info] SUITE ABORTED - CassandraSparkSpec: Failed to open native connection to Cassandra at {0.0.0.0}:9042
[info]   java.io.IOException: Failed to open native connection to Cassandra at {0.0.0.0}:9042
```

bzz:
@dpordomingo thank you for double-checking!

The only thing is, while it works locally, somehow it can not connect to DB in container at `0.0.0.0:9042` while on CI :/

smacker:
it's bad idea to use `0.0.0.0` when you are connecting to something. In terms of listening it usually means all interfaces. But in case of connecting it usually means invalid address. Depends on system configuration. Anyway I would recommend you to connect to 127.0.0.1 instead.

smacker:
you may want to look at this small utility for waiting: https://github.com/jwilder/dockerize

bzz:
@smacker thanks for the heads up! Used a simpler https://github.com/vishnubob/wait-for-it

CI would pass now, sans #4 

Merging, and moving to #4 if there is no further discussion.

dpordomingo:
LGTM

PR #10, opened 2017-12-11
Query/Hash: apply schema

### LINKED ISSUE #4: Make all commands apply schema (if not exist)

bzz:
This PR adds
 - [x] lazy schema creation for Query and Hash commands
 - [x] fixes some build irregularities

All this enables the CI to pass and closes #4

### DIFF
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -36,13 +36,17 @@ class Gemini(session: SparkSession) {
     filesToWrite
   }
 
-  def hashAndSave(reposPath: String): Unit = {
-    val files = hash(reposPath)
+  def save(files: DataFrame): Unit = {
     println(s"Writing ${files.rdd.countApprox(10000L)} files to DB")
     files.write
       .mode("append")
       .cassandraFormat("blob_hash_files", "hashes")
       .save()
+  }
+
+  def hashAndSave(reposPath: String): Unit = {
+    val files = hash(reposPath)
+    save(files)
   }
 
 }
@@ -95,16 +99,10 @@ object Gemini {
 
   def findDuplicateFiles(file: File, conn: Session): Iterable[RepoFile] = {
     val sha = computeSha1(file)
-
     val query = QueryBuilder.select().all().from("hashes", "blob_hash_files").where(QueryBuilder.eq("blob_hash", sha))
-    var results: Iterable[RepoFile] = Iterable()
-
-    results = conn
-      .execute(query)
-      .asScala map { row =>
-        RepoFile(row.getString("repo"), row.getString("file_path"), row.getString("blob_hash"))
-      }
-    results
+    conn.execute(query).asScala.map { row =>
+      RepoFile(row.getString("repo"), row.getString("file_path"), row.getString("blob_hash"))
+    }
   }
 
   def computeSha1(file: File): String = {
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -19,7 +19,10 @@ class Gemini(session: SparkSession) {
   def hash(reposPath: String): DataFrame = {
     val engine = Engine(session, reposPath)
 
-    val headRefs = engine.getRepositories.getHEAD.withColumnRenamed("hash", "commit_hash")
+    val headRefs = engine.getRepositories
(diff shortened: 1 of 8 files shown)


bzz:
CI is not green for now presumably due to https://github.com/src-d/engine/issues/255

Will switch to simpler case of using a .sive file instead of git repo.

bzz:
CI is green, merging if there is no discussion.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -32,13 +32,24 @@ class Gemini(session: SparkSession) {
 
     filesToWrite
   }
+
+  def hashAndSave(reposPath: String): Unit = {
+    val files = hash(reposPath)
+    println(s"Writing ${files.rdd.countApprox(10000L)} files to DB")
+    files.write
+      .mode("append")
```
smacker: not really related to pr, just for my understanding.
Could you please explain what would happen if I run `hashAndSave` 2 times for the same repo? I remember you said primary key isn't unique in cassandra. Will it duplicate hashes/files? Can it be a problem later when we query db?

bzz: Good question.

1. It is very easy to try and see what will actually happen when you run this code. 
    Here are the detailed steps from https://github.com/src-d/gemini#pre-requests on running analog of the [integration test](https://github.com/src-d/gemini/pull/10/files#diff-354f30a63fb0907d4ad57269548329e3R22) locally for your convenience:
   ```
   $ docker run -p 9042:9042 \
    --name some-scylla -d scylladb/scylla:2.0.0 \
    --broadcast-address 127.0.0.1 --listen-address 0.0.0.0 --broadcast-rpc-address 127.0.0.1 \
    --memory 2G --smp 1

   $ git clone git@github.com:src-d/gemini.git; cd gemini
   $ git fetch origin pull/10/head:fix-ci-schema

   $ ./hash src/test/resources/siva

   $ docker exec -it some-scylla cqlsh

    # in CQLsh
    select count(*) from blob_hash_files;

    # then hash the same repository again
    ./hash src/test/resources/siva 

    # again, in CQLsh
    select count(*) from blob_hash_files;
   ```
2. As we discussed, [PRIMARY KEY](http://docs.scylladb.com/getting-started/ddl/#the-primary-key) concept is indeed different in C* then in RDBMS. One have to think about C* as a distributed KV storage with K space being sorted, [partitioned](http://docs.scylladb.com/getting-started/ddl/#the-partition-key) and [clustered](http://docs.scylladb.com/getting-started/ddl/#the-clustering-columns). I.e if clustering is not defined in PR, then partition will always consist of a single row and guarantee uniqnes, similar to RDBMS.

In case at hand we have a PK statement in [schema](https://github.com/src-d/gemini/blob/master/src/main/resources/schema.cql#L3) that defines both, partitioning and clustering keys, and so entities with same hashes will end up on a single partition that would consist of a set of rows, sorted by repository.

smacker: haha. I don't have enough RAM to run it :-D

I also noticed then that `file_path` may need to be `file_paths` in schema.
Repository may contain 2 or more identical files. The `blob_hash` will be the same for them.

bzz: Thanks for double-check! 
`file_paths` might lead to the impression that the value has a type of collection or map inside the table so I avoid using plurals for that reason.

smacker: hm. My idea is we may need collection there. Currently for copy-pasted files:
```
/src/util.go
/src/some/package/cool_util.go
```
in the same repository we will have only one and duplication won't appear in report. Maybe we want to change it. Just a food for thoughts.

bzz: Hm, not sure I understand why is this is going to happen, could you elaborate?

The best would be if you could reproduce this locally using this branch, with adding a copy of the file in cloned repo,`./hash </path-to-cloned-repo-parrent/>` and  `./query <copy of the file>`.

smacker: Because key is the same for them. Created issue: https://github.com/src-d/gemini/issues/15

bzz:
@dpordomingo thank you for the feedback! addressed in b9d38fe

bzz:
Merging if there is no further discussion

Issue #8, opened 2017-12-11
Implement report command


dpordomingo:
This command will output the duplicate files among all the hashed files:
- [x] `<file_path> | <repo> | <sha1>`
- [ ] show the `<sha1>` only when command is ran with `--verbose` mode
- [ ] add a lik to the file in GH (based in HEAD) -> `https://github.com/<repo>/blob/HEAD/<file_path>`

dpordomingo:
I updated the issue description considering the feedback that @bzz gave by Slack

PR #11, opened 2017-12-11
Find duplicated files

dpordomingo:
fix https://github.com/src-d/gemini/issues/8

Implements the `report`.
The default behavior does not use `GROUP BY` query, so it is Cassandra and Scylla compatible, and performs one query per distinct file.

If it's used a Cassandra database, there can be used two different flags:
- `--condensed` returns the duplicated `blob_hash` and how many duplicates exists for each one —with only one query—,
- `--use-group-by`, same output than the default one (no flags), but using `GROUP BY` —with `1 + n` queries, where `n` is the number of unique `blob_hash` of duplicate files—

example:

```shell
> ./report --condensed
Duplicates found:
	493d45ba4391e78dbc258934c9ecebabd5ae8d40 (2 duplicates)
	dc6817245a5213ab250471276acb61c186acab11 (2 duplicates)
	44f67e8d1d20f512b46bd9fb395a8c72982ffaa0 (2 duplicates)
	2fa601c8c29c75c7b5d7ec80a22d5dbfde82dafc (2 duplicates)
	39c18589882021fabe95452d7a2fbf3bdcc9a95d (2 duplicates)
	0ee906abfeffb8e2ed1751ba1113fa71bb9d79da (2 duplicates)
	faa0e8f5e75c41b11bdbfbf439456ffbf94c727e (2 duplicates)
```

```shell
> ./report
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,model_test.go,493d45ba4391e78dbc258934c9ecebabd5ae8d40)
	RepoFile(github.com/src-d/borges.git,model_test.go,493d45ba4391e78dbc258934c9ecebabd5ae8d40)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,MAINTAINERS,dc6817245a5213ab250471276acb61c186acab11)
	RepoFile(github.com/src-d/borges.git,MAINTAINERS,dc6817245a5213ab250471276acb61c186acab11)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,changes.go,44f67e8d1d20f512b46bd9fb395a8c72982ffaa0)
	RepoFile(github.com/src-d/borges.git,changes.go,44f67e8d1d20f512b46bd9fb395a8c72982ffaa0)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,model.go,2fa601c8c29c75c7b5d7ec80a22d5dbfde82dafc)
	RepoFile(github.com/src-d/borges.git,model.go,2fa601c8c29c75c7b5d7ec80a22d5dbfde82dafc)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,cli/borges/version.go,39c18589882021fabe95452d7a2fbf3bdcc9a95d)
	RepoFile(github.com/src-d/borges.git,cli/borges/version.go,39c18589882021fabe95452d7a2fbf3bdcc9a95d)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,Makefile,0ee906abfeffb8e2ed1751ba1113fa71bb9d79da)
	RepoFile(github.com/src-d/borges.git,Makefile,0ee906abfeffb8e2ed1751ba1113fa71bb9d79da)
2 duplicates:
	RepoFile(github.com/erizocosmico/borges.git,doc.go,faa0e8f5e75c41b11bdbfbf439456ffbf94c727e)
	RepoFile(github.com/src-d/borges.git,doc.go,faa0e8f5e75c41b11bdbfbf439456ffbf94c727e)
```


dpordomingo:
I'd appreciate if you take a look to the PR to ensure the output format and global style is ok.
If you agree with current state, I'll continue (before merging) with:
- [ ] tests, of course,
- [ ] `--no-group-by` flag that will force the command to make a single query per file, to obtain the duplicates

dpordomingo:
CI won't pass due to https://github.com/src-d/gemini/issues/4
I'll rebase tomorrow over https://github.com/src-d/gemini/pull/10 to make the CI pass

bzz:
@dpordomingo the general approach looks good to me!

Could you also make sure you apply InteliJ Auto-formatter to the code?

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -58,30 +61,52 @@ object Gemini {
...
-    throw new UnsupportedOperationException("Finding all duplicate files in many repositories is no implemented yet.")
+  def detailedReport(conn: Session): Iterable[Iterable[RepoFile]] = {
+    findDuplicateBlobHash(conn)
+        .map {d => findDuplicateBlobHash(d.sha, conn)}
+  }
+
+  def findDuplicateBlobHash(conn: Session): Iterable[DuplicateBlobHash] = {
+    val cql = "SELECT blob_hash, repo, file_path, COUNT(*) as count FROM hashes.blob_hash_files GROUP BY blob_hash"
+    conn
+      .execute(new SimpleStatement(cql))
```
bzz: usually it's a good idea to extract constants with the queries as a top level fields in object/class.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -58,30 +61,52 @@ object Gemini {
...
-    * @param repository repository url, example: github.com/src-d/go-git.git"
+    * @param conn   Database connection
     * @return
     */
-  def report(repository: String): Iterable[RepoFile] = {
-    throw new UnsupportedOperationException("Finding all duplicate files in many repositories is no implemented yet.")
+  def detailedReport(conn: Session): Iterable[Iterable[RepoFile]] = {
+    findDuplicateBlobHash(conn)
+        .map {d => findDuplicateBlobHash(d.sha, conn)}
+  }
```
bzz: this can be re-formatted a bit

```scala
=  findAllDuplicateItems(conn).map { item => 
          findDuplicateHashesFor(item.sha, conn)
   }

```

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -58,30 +61,52 @@ object Gemini {
...
     throw new UnsupportedOperationException("Finding duplicate repositories is no implemented yet.")
   }
 
-  def findDuplicateFiles(file: File, conn: Session): Iterable[RepoFile] = {
-    val sha = computeSha1(file)
+  def findDuplicateBlobHash(file: File, conn: Session): Iterable[RepoFile] = {
+    findDuplicateBlobHash(computeSha1(file), conn)
+  }
 
+  def findDuplicateBlobHash(sha: String, conn: Session): Iterable[RepoFile] = {
```
bzz: Having two overriden methods with the same names that have only different argument types usually is not a good idea - it does not help a reader to understand the purpose or difference of those methods.

One suggesion from above is to have `findAllDuplicateItems` for listing the dups and then `findDuplicateHashesFor` a given item/blobHash

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,49 @@
...
+  val session = cluster.connect()
 
-  //for every file
-  // query Cassandry by hash
-  // .collect()
-  //print
+  if (detailed) {
+    val report = Gemini.detailedReport(session)
+    printIsEmpty(report)
+    report.foreach( duplicateFiles => {
```
bzz: this is the case, when I wold prefer more concice

```scala
.foreach { duplicateFiles =>
}
```
form

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,49 @@
...
+      val count = duplicateFiles.count(_=>true)
+      println (s"$count duplicates:\n\t" + (duplicateFiles mkString ("\n\t")))
+    })
+  } else {
+    val report = Gemini.report(session)
+    printIsEmpty(report)
```
bzz: These two lines look very similar in both cases of `if`, so it may be it could better to move them out of `if` with a flag like  `Gemini.report(session, detailed)` and keep `if` here only for conditional result display?

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,49 @@
...
     System.exit(2)
   }
 
-  if (args.length <= 0) {
-    printUsage()
+  def printIsEmpty[T](report:Iterable[T]): Unit = {
+    if (report.isEmpty) {
+      println(s"No duplicates found.")
+    }
   }
```
bzz:
```scala
def printIsEmpty(report: Traversable[Any]): Unit = if (report.isEmpty) {
    println(s"No duplicates found.")
}
```

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: Andas, many thanks for the hint :)

dpordomingo:
@bzz you can review the next two commits, that:
- address your suggestions 3bb8122
- adds tests 5bb3c77

Missing things:
- [ ] implement `Gemini.applySchema` from https://github.com/src-d/gemini/pull/10/files
- [ ] `--no-group-by` flag that will force the command to make a single query per file, to obtain the duplicates

bzz:
@dpordomingo awesome, let me make another pass on it before merging!

From a quick glance - 👍 for having tests!

Meanwhile, in this PR let's: 
 - make sure CI is green (it's good practice to post the reason of CI failure in the issue)
 - add `Gemini.applySchema`
 - make sure report is called as part of our Integration end-to-end test on CI 

And make `--no-group-by` in the next one, on top of #16 ?

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -76,29 +82,54 @@ object Gemini {
...
     * @return
     */
-  def report(repository: String): Iterable[RepoFile] = {
-    throw new UnsupportedOperationException("Finding all duplicate files in many repositories is no implemented yet.")
+  def findAllDuplicateItems(conn: Session): Iterable[DuplicateBlobHash] = {
+    conn
+      .execute(new SimpleStatement(cql))
+      .asScala
+      .filter(_.getLong("count") > 1)
+      .map { r => DuplicateBlobHash(r.getString("blob_hash"), r.getLong("count")) }
```
bzz: would you prefer keeping it like this, rather then

```scala
.map { r => 
  DuplicateBlobHash(r.getString("blob_hash"), r.getLong("count"))
}
```
?

It's not strong preference, this on just have advantage of being simpler to pattern-match by human eye.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -53,11 +53,17 @@ class Gemini(session: SparkSession) {
...
+case class DuplicateBlobHash(sha: String, count: Long) {
+  override def toString(): String = s"$sha ($count duplicates)"
+}
+
 object Gemini {
   val defaultCassandraHost: String = "127.0.0.1"
   val defaultCassandraPort: String = "9042"
   val defaultSchemaFile: String = "src/main/resources/schema.cql"
 
+  val cql = "SELECT blob_hash, repo, file_path, COUNT(*) as count FROM hashes.blob_hash_files GROUP BY blob_hash"
```
smacker: do you think `duplicatesCountCql` or something like this would be better?

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,46 @@
...
+    println("  --detailed to see file paths of each duplicated file")
     System.exit(2)
   }
 
-  if (args.length <= 0) {
-    printUsage()
+  def print(report: Iterable[Any], detailed: Boolean): Unit = {
+    if (report.isEmpty) {
+      println(s"No duplicates found.")
+    } else if (detailed) {
```
smacker: is early return common practice in Scala?

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,46 @@
...
+      report.foreach { item =>
+        val duplicateFiles = item.asInstanceOf[Iterable[RepoFile]]
+        val count = duplicateFiles.count(_ => true)
```
smacker: can you use `duplicateFiles.size` here?

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: why do you need to select `repo` and `file_path` here if you are grouping anyway?

smacker: actually it's weird. I don't see support of GROUP BY in cassandra. How does it work?

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
bzz: Usually it is not.
If you have concrete suggestions on restructuring the code here - please feel free to share examples as well.

src/main/scala/tech/sourced/gemini/Gemini.scala
bzz: @smacker it's rather new https://github.com/scylladb/scylla/issues/2206
and that is why this PR has 2 implementations: with and without it.

smacker: So I can't use it right now with scylla? :(

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
smacker: Regular early return as in go or js or many other languages:
```scala
def print(report: Iterable[Any], detailed: Boolean): Unit = {
    if (report.isEmpty) {
        println(s"No duplicates found.")
        return
    }

    if (detailed) {
        report.foreach { item =>
        val duplicateFiles = item.asInstanceOf[Iterable[RepoFile]]
        val count = duplicateFiles.count(_ => true)
        println(s"$count duplicates:\n\t" + (duplicateFiles mkString "\n\t") + "\n")
        return
    } 

    println(s"Duplicates found:\n\t" + (report mkString "\n\t"))
}
```

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
dpordomingo: It's great to know that `.count(_=>true)` method is not the proper way to do it.
Thanks for shaming me. I really deserved it :dagger:

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: In Cassandra since more than 1y ago. There is not in Scylla :(

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: I'll try to apply your suggestion as a regular pattern in iterations while they're more than 80chars length. Thanks for noticing :dagger:

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
smacker: haha. no shaming man. we are learning. Hope you will point my mistakes too. :)

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
bzz: Yes, and that is not common practice and generally shall be avoided in Scala. 
Here is more context https://tpolecat.github.io/2014/05/09/return.html

src/main/scala/tech/sourced/gemini/Gemini.scala
bzz: Yes. Let's make a no-group-by a default behaviour for now

src/main/scala/tech/sourced/gemini/Gemini.scala
bzz: I see! Two things
 - it may be a good idea to use this pattern to assist readably regardless of the length of the line
 - in this project, max line length is [120 chars](https://github.com/src-d/gemini/blob/master/scalastyle-config.xml#L17) which is friendlier to modern monitors

bzz:
I guess that current status is:

TODO:
 - refactoring to add idempotent tests that use different keyspace
 - incorporate review feedback
 - make no `no-group` default impl

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
smacker: Oh I see why now. Thanks!

Actually it makes sense, you always can use pattern matching.

```scala
sealed abstract class Result {
    def size(): Int
}
case class ResultFiles(vs: List[RepoFile]) extends Result {
    override def size() = vs.size
}
case class ResultBlobs(vs: List[DuplicateBlobHash]) extends Result {
    override def size() = vs.size
}

def print(report: Result): Unit = {
    report match {
        case empty if empty.size == 0 => println(s"No duplicates found.")
        case ResultFiles(vs) => println(vs.size+" duplicates:\n\t" + (vs mkString "\n\t") + "\n")
        case ResultBlobs(vs) => println(s"Duplicates found:\n\t" + (vs mkString "\n\t"))
    }
}
```

smacker: not need to rewrite though. Just example of cool scala syntax.

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: I suggested the 80chars max for iterator callbacks to avoid the horizontal scroll in the GitHub interface; anyhow, I can take that rule for all callback iterators regardless its length.

dpordomingo:
- [x] implement `Gemini.applySchema` for `report` command from PR #10 846ab0e
- [x] make the tests to be idempotent, using different keyspaces 846ab0e
- [x] incorporate review feedback 846ab0e
- [x] call `./report` as part of our [Integration end-to-end test on CI](https://github.com/src-d/gemini/blob/master/.travis.yml#L27) 16285e0
- [ ] make sure CI is green
- [ ] make the report command to run a single query per file, to obtain the duplicates, and use `--group-by` flag to fetch the duplicate files with only one `GROUP BY` query

I think CI is failing because `scylladb/scylla:2.0.0` does not support `GROUP BY`. I wanted to do it in a separated PR, but if so, built will not pass.

To be done in a separate PR
- [ ] make sure the commands fail gracefully with a good user feedback in case if no DB.
.

dpordomingo:
@bzz all comments were addressed, pending sub-tasks were added and CI now passes. :fireworks: :star2:

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -49,81 +52,182 @@ class Gemini(session: SparkSession) {
...
+
+  def applySchema(session: Session): Unit = {
+    println("CQL: creating schema") //TODO(bzz): Log.Debug
+    Source
+      .fromFile(Gemini.defaultSchemaFile)
+      .getLines
+      .map(_.trim)
+      .filter(!_.isEmpty)
+      .foreach { line =>
+        val cql = line.replace("__KEYSPACE__", keyspace)
```
smacker: is it possible to just execute `USE $keyspace` before CQL and remove prefix in cql file?

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -49,81 +52,182 @@ class Gemini(session: SparkSession) {
...
-        session.execute(line)
-      }
-    println("CQL: Done. Schema created")
+sealed abstract class Report(v: Iterable[Any]) {
+  def empty(): Boolean = {
+    v.isEmpty
   }
 
+  def size(): Int = {
+    v.asInstanceOf[Iterable[Any]].size
```
smacker: it should be possible to simplify as:
```
def size(): Int = v.size
```

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
```diff
@@ -1,34 +1,67 @@
...
+      "\n\tIt uses only one query to find the duplicates. (only for Apache Cassandra database)"
+    )
     System.exit(2)
   }
 
-  if (args.length <= 0) {
+  def print(report: Report): Unit = {
+    report match {
+      case e if e.empty() => println(s"No duplicates found.")
+      case r: ReportGrouped => println(s"Duplicates found:\n\t" + (r.v mkString "\n\t"))
```
smacker: I'm only learning Scala, but I saw class matching for such cases very often in examples.
I mean
```
case ReportGrouped(v)=> print(v)
```
JFYI.

src/main/scala/tech/sourced/gemini/Gemini.scala
bzz: It could be one possible further improvement. 
For now, this replacement already looks awesome.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -49,81 +52,182 @@ class Gemini(session: SparkSession) {
...
     * @return
     */
-  def report(repository: String): Iterable[RepoFile] = {
-    throw new UnsupportedOperationException("Finding all duplicate files in many repositories is no implemented yet.")
+  def findAllDuplicateItems(conn: Session, keyspace: String): Iterable[Iterable[RepoFile]] = {
+    val distinctBlobHash = s"SELECT distinct blob_hash FROM ${keyspace}.blob_hash_files"
+    conn
+      .execute(new SimpleStatement(distinctBlobHash))
+      .asScala
+      .flatMap { r => {
```
bzz: To be consistent, let's remove the second `{`

.travis.yml
```diff
@@ -16,7 +16,7 @@ matrix:
   - scala: 2.11.2
     script:
       - ./scripts/wait-for-it.sh 127.0.0.1:9042 -- echo "Cassandra/ScyllaDB is up"
-      - ./sbt test
+      - ./sbt "test-only -- -l Cassandra"
```
bzz: Why do you think this is required?

Having sub-sets of test labeled is very nice (esp. to filter out those big ones, that has external dependencies), but for this CI profile, it seems that it most probably would make sense to execute all of them. 

This has additional benefit of being a very simple `./sbt test` which is quite standard.

dpordomingo: The tests-cases labeled as "Cassandra" requires Apache Cassandra, not available in the CI (because we use Scylla)

If we could use two different database images with Travis, we could use the Cassandra one to pass these tests (but I think it should be done in a separated PR)

src/main/scala/tech/sourced/gemini/ReportSparkApp.scala
dpordomingo: many thanks for the advice!

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: great! thanks!

src/main/scala/tech/sourced/gemini/Gemini.scala
dpordomingo: I'm not sure if it could cause unexpected problems. Could it happen something weird if running parallel tests with a shared connection, or would it be forbidden?

bzz:
One thing I noticed - on `./report --condensed` throws an exception for me: `com.datastax.driver.core.exceptions.SyntaxError: line 1:23 no viable alternative at input '('
`

Would make sense to add it to end-to-end CI profile as well?


<details>

```
[error] (run-main-0) com.datastax.driver.core.exceptions.SyntaxError: line 1:23 no viable alternative at input '('
com.datastax.driver.core.exceptions.SyntaxError: line 1:23 no viable alternative at input '('
	at com.datastax.driver.core.exceptions.SyntaxError.copy(SyntaxError.java:58)
	at com.datastax.driver.core.exceptions.SyntaxError.copy(SyntaxError.java:24)
	at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:37)
	at com.datastax.driver.core.DefaultResultSetFuture.getUninterruptibly(DefaultResultSetFuture.java:245)
	at com.datastax.driver.core.AbstractSession.execute(AbstractSession.java:68)
	at tech.sourced.gemini.Gemini$.findAllDuplicateBlobHashes(Gemini.scala:169)
	at tech.sourced.gemini.Gemini.reportCassandraCondensed(Gemini.scala:95)
	at tech.sourced.gemini.ReportSparkApp$.delayedEndpoint$tech$sourced$gemini$ReportSparkApp$1(ReportSparkApp.scala:59)
	at tech.sourced.gemini.ReportSparkApp$delayedInit$body.apply(ReportSparkApp.scala:5)
	at scala.Function0$class.apply$mcV$sp(Function0.scala:34)
	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
	at scala.App$$anonfun$main$1.apply(App.scala:76)
	at scala.App$$anonfun$main$1.apply(App.scala:76)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
	at scala.App$class.main(App.scala:76)
	at tech.sourced.gemini.ReportSparkApp$.main(ReportSparkApp.scala:5)
	at tech.sourced.gemini.ReportSparkApp.main(ReportSparkApp.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
Caused by: com.datastax.driver.core.exceptions.SyntaxError: line 1:23 no viable alternative at input '('
	at com.datastax.driver.core.Responses$Error.asException(Responses.java:132)
	at com.datastax.driver.core.DefaultResultSetFuture.onSet(DefaultResultSetFuture.java:179)
	at com.datastax.driver.core.RequestHandler.setFinalResult(RequestHandler.java:177)
	at com.datastax.driver.core.RequestHandler.access$2500(RequestHandler.java:46)
	at com.datastax.driver.core.RequestHandler$SpeculativeExecution.setFinalResult(RequestHandler.java:799)
	at com.datastax.driver.core.RequestHandler$SpeculativeExecution.onSet(RequestHandler.java:633)
```

</details>

.travis.yml
bzz: Thank you for explanation! If it excludes a Cassandra ones - that sounds good!

For now, we do not want to support 2 DB, so the solution looks good.

dpordomingo:
If you're using Scylla or any other db non supporting `GROUP BY`, it is expected a syntax error.
We should prepare the command for that "expected exceptions", but since it's already documented, I think we could do it in a separated PR.

.travis.yml
dpordomingo: :+1: You can find that changes (excluding the Cassandra tagged test cases) in this commit https://github.com/src-d/gemini/pull/11/commits/0c661c6b9952e398892443b732976149d4f2b48a

dpordomingo:
Both, `--condensed` and `--use-group-by` requires Cassandra and uses `GROUP BY` query.
I agree,  we should improve the wording and the usage of flags

smacker:
Looks like everyone agreed to merge it. So I'm merging to rebase my PR on top of it.

PR #13, opened 2017-12-12
ScyllaDB deployment for kubernetes

smacker:
For now supports only one instance in cluster.
Tested on minikube.

### DIFF
--- a/k8s/scylla-statefulset.yaml
+++ b/k8s/scylla-statefulset.yaml
@@ -55,7 +55,10 @@ spec:
             mountPath: /opt/ready-probe.sh
             subPath: ready-probe.sh
       nodeSelector:
-        srcd.host/type: worker-applications
+        # selector should be:
+        # srcd.host/type: worker-applications
+        # but for now let's pin it to particular node
+        kubernetes.io/hostname: worker-4.k8s.pipeline.staging.srcd.host
       volumes:
         - name: scylla-data # don't forget to change to persistent volume
           emptyDir:
--- a/k8s/scylla-statefulset.yaml
+++ b/k8s/scylla-statefulset.yaml
@@ -1,0 +1,65 @@
+apiVersion: apps/v1beta2
+kind: StatefulSet
+metadata:
+  name: scylladb
+  labels:
+    app: scylladb
+spec:
+  serviceName: scylladb
+  replicas: 1
+  selector:
+    matchLabels:
+      app: scylladb
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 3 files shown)


bzz:
Did not try it yet with kubectl.
\cc @rporres for a quick glance, in case of some spare cycles.

k8s/scylla-statefulset.yaml
```diff
@@ -0,0 +1,65 @@
+apiVersion: apps/v1beta2
```
rporres: I think k8s 1.5 does not support this apiVersion

k8s/scylla-statefulset.yaml
```diff
@@ -0,0 +1,65 @@
...
+          initialDelaySeconds: 15
+          timeoutSeconds: 5
+        volumeMounts:
+          - name: scylla-data
+            mountPath: /var/lib/scylla
+          - name: scylla-ready-probe
+            mountPath: /opt/ready-probe.sh
+            subPath: ready-probe.sh
+      nodeSelector:
+        # add selecter here when node is ready
```
rporres: `srcd.host/type: worker-applications`

rporres: Or if you want something more restrictive, you can use `hostname: worker-4.k8s.pipeline.staging.srcd.host`

smacker:
Thanks! Version changed, selector added.
I thought to create helm chart, but Alex said just such config is enough. We will migrate it to helm later.

rporres:
If you're going to put the database in disk, most likely you want to tie your stateful set to a host rather than a set of hosts as there are three `srcd.host/type: worker-applications`. If not, it's ok

smacker:
Yeah, thanks. I'll discuss it with Alex. Normally we would need persistent volume, but for testing just mounted directory should be fine.

Tried to deploy, it works:
```
smacker at Maxims-MacBook-Air in ~/srcd/gemini on scylla_db_k8s*
$ kubectl exec -it rafa-test-3698564540-pzn0r sh
/ # telnet scylladb 9042

�1
+Invalid or unsupported protocol version: 13Connection closed by foreign host
/ #
```

P.S. sorry for using your pod :-D

bzz:
@rporres thank you for as usually thoughtful feedback!

@smacker nice! BTW what is `DCO_2` file for?

smacker:
oh. sorry. Haven't noticed test commit. Removed.

smacker:
@bzz is anything else required here or we can merge?

PR #16, opened 2017-12-13
add proper CLI and parametarise connection

smacker:
Fixes: #2 #12

@bzz please let me know how it should be written in scala. I don't know is huge pattern matching okay here or not.

Also I tried to avoid copy-pasting of db options, but didn't find good solution: https://github.com/scopt/scopt/issues/132
there are _some_ solutions but none of them I like.

PR #17, opened 2017-12-19
Pass arguments as a string instead of as an array

dpordomingo:
I realized that `report` fails when args are passed as an array (`$@`) because `exec` needs a string of args
Using `$*` the args are passed as a string

### DIFF
--- a/query
+++ b/query
@@ -18,5 +18,5 @@ if [[ ! -f "${jar}" ]]; then
 fi
 
 
-#exec java -cp "${jar}" "${app_class}" $@
-exec ./sbt "run-main tech.sourced.gemini.QueryApp $@"
+#exec java -cp "${jar}" "${app_class}" $*
+exec ./sbt "run-main tech.sourced.gemini.QueryApp $*"
--- a/report
+++ b/report
@@ -17,4 +17,4 @@ if [[ ! -f "${jar}" ]]; then
     fi
 fi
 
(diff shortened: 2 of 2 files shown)


smacker:
https://github.com/src-d/gemini/pull/16#discussion_r157274107

dpordomingo:
@smacker done; thanks for the review :dagger:

smacker:
maybe you looked at the wrong comment. I meant this one from Alex:

> Got it! Thank you for detailed example.
> 
> Two things:
> 
> in general, this is anit-pattern and you almost always want $@ - I find this shell styleguide to have a good explanation on such topics
> in this case, indeed, the way non-Spark app .sh wrappers work with ./sbt "run-main ..." does require this, as a workaround. Good catch!
> Let's limit this change to the places where its required.

dpordomingo:
If @bzz agree, we can merge it

bzz:
I'll go though it today and get back to you.

bzz:
I believe, it's already been addressed in https://github.com/src-d/gemini/pull/16/ so may be we could keep it that way and close this in favor of #16 ?

AFAIK, as noted in comment cited by @smacker,  it's not exec who needs it, but rather `./sbt "run-main"`

smacker:
It was resolved in https://github.com/src-d/gemini/pull/16

Issue #21, opened 2017-12-20
Update output of Spark application


bzz:
For `./hash`
 migrate to [logger](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/Logging.scala), provided by Apache Spark
 output: mute meaningless Spark logs (though log4j.config)
 

This would most probably mean to refactor `Gemini` class to have logging dependency injected (thorough constructor or `.set...()`), as it will be instantiated by the client (Spark/non-spark ones) in a different ways.

smacker:
fixed

Issue #19, opened 2017-12-20
Update output of scala applications


bzz:
For `query` and `report` tools:
 -  add -v flag, producing more verbose debug output
 - replace `println()` with proper slf4j logging (same as Apache Spark uses) on different levels INFO/DEBUG

The Spark Application logging will be handled in different issue.

smacker:
fixed

PR #24, opened 2017-12-22
Use logger for scala output

smacker:
This PR fixes both https://github.com/src-d/gemini/issues/19 and https://github.com/src-d/gemini/issues/21 because gemini uses the same logger as spark. So fixing one issue fixes another too.

Logging in java is complicated at first.

So what I have done here:

1. Gemini class accepts `org.slf4j.Logger` now. And print information using `log.debug`. `org.slf4j.Logger` is commonly used interface for logging. It doesn't depend on implementation like`log4j` or `logback` or whatever else.
2. Added `log4j12` as a dependency, because spark uses it. Documentation of log4j by default points to a new version of `log4j` so it's a bit confusing. Be aware!
3. Added `slf4j-log4j12` as a dependency. This package is a bridge between`slf4j` API and `log4j12` implementation.
4. Simplified configuration in `log4j.properties`. By default, it prints only errors.
5. Added `verbose` flag which sets INFO level for root logger. It means all dependency will print INFO messages.
6. It's possible to enable DEBUG logging for root logger or any other configuration by passing `-Dlog4j.configuration=path/to/log4j.properties` to java.

Example of report command (sbt logging is stripped here and in the next listings):
https://gist.github.com/smacker/182d88934b894f52b49cf0aecdc1d8e9

Report command with `--verbose` flag:
https://gist.github.com/smacker/e6e8086b8f50e300ab2e509664e47b18

Hash command:
```
Hashing 3 repositories in: src/test/resources/siva/
	file:/Users/smacker/Work/gemini/src/test/resources/siva/duplicate-files
	file:/Users/smacker/Work/gemini/src/test/resources/siva/unique-files
	file:/Users/smacker/Work/gemini/src/test/resources/siva/2636f3c62f1a407b2996da6e3fe6fdc5d1ccd764.siva
Done <- while processing it shows progress here
```

Hash command with `--verbose` flag:
https://gist.github.com/smacker/2006ac4bcbdbe4b8c8d9dd0c6e08e9ab

Please let me know if you think defaults should be different. Or something inside gemini should be info instead of debug.


src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -54,7 +55,7 @@ class Gemini(session: SparkSession, keyspace: String = Gemini.defautKeyspace) {
   }
 
   def save(files: DataFrame): Unit = {
-    println(s"Writing ${files.rdd.countApprox(10000L)} files to DB")
+    log.debug(s"Writing ${files.rdd.countApprox(10000L)} files to DB")
```
dpordomingo: Imho this would be useful as a .`info`

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -161,7 +162,9 @@ object Gemini {
 
   val formatter = new ObjectInserter.Formatter
 
-  def apply(ss: SparkSession, keyspace: String = defautKeyspace): Gemini = new Gemini(ss, keyspace)
+  def apply(ss: SparkSession, log: Slf4jLogger = Logger("gemini"), keyspace: String = defautKeyspace): Gemini = {
```
dpordomingo: `{` and `}` is not needed

smacker:
The idea to add logger in constructor was mentioned in the issue.
DI sounds good but I don't know how to do it "the right way" in Scala.
I found this article: http://jonasboner.com/real-world-scala-dependency-injection-di/
In scala you can do DI without any external libs and it's nice! But I'm not sure it's a really correct way.

I would suggest for now to keep it in constructor as it was described in the issue. And later discuss DI solution. Most probably Alex knows about it more.

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: I had doubts about this one. Thought it's more like debug information, but now I have changed my mind. Fixed!

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: Fixed

dpordomingo:
Good idea, let's talk about DI once we're there

LGTM

bzz:
Looks great! Few things below. 

Do you think it could be possible to :
 1. make `./sbt test` less verbose by default
 2. for the case of `HashingSparkApp` - instead of creating logger manually, to use one instantiate by [Spark logging](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/Logging.scala) facilities with something like 
    ```scala
    object HashSparkApp extends App with Logging
    ```

smacker:
@bzz 
1. How much less verbose do you want it to be? It already has the custom configuration. WARN for all, INFO for datastax and ERROR level for many other things. Do you want to set ERROR for everything? Or WARN? Or something more custom?
2. Don't really understand what you mean. Spark's Logging trait will add logging methods to the app, but it won't allow changing logging level programmatically.

bzz:
1. Just saying that when `./sbt test` is run - I do not see a reason right now to have looooong meaningless output. I believe reasonable default experience would be not to show all the ` INFO 15:24:41,548 com.datastax.spark.connector.util.Logging$class (Logging.scala:35) - Wrote 0 rows to test_hashes_uniques.blob_ha...`

2. Sorry for confusion. Instead of https://github.com/src-d/gemini/pull/24/files#diff-157389cdfb1cdd4e84efe27f02ae34efR44 can we use the logger instance, created by Spark logging? It has advantage of taking care of logging in a distributed environment.

   Being able to use either or them inside Gemini interchangeably - that's what DI in https://github.com/src-d/gemini/issues/21 was supposed to mean.

smacker:
1. Okay then, I have just change level of datastax to WARN
2. Got your point now. Thanks! Fixed.

Issue #9, opened 2017-12-11
Add URLs to the output of report and query commands


dpordomingo:
Show links to the files when it's found duplicates.

- [x] Add reference hash to DB.
- [x] For Github `https://github.com/<repo>/blob/<ref_hash>/<file_path>`
- [x] For Gitlab - same URL schema is the same as github
- [x] For BitBucket, according to [docs](https://confluence.atlassian.com/bitbucket/hyperlinking-to-source-code-in-bitbucket-824476709.html) ` https://bitbucket.org/<repo>/src/<ref_hash>/<file_path>`

marnovo:
Are the "links" targeting a remote or local address? Could you elaborate a bit? Thanks!

dpordomingo:
The links would point to the GitHub too if the files come from there
`https://github.com/<repo>/blob/HEAD/<file_path>`
If the files come from any other source, it should be analyzed separately

bzz:
Updated issue description \w details on Bitbucket

smacker:
there is also gitlab as I mentioned before. It uses the same url schema as github, but different domain.

bzz:
> there is also gitlab 

BTW, from what I can see https://github.com/src-d/rovers/tree/master/providers there might not be gitlab.com support across the stack yet.

smacker:
but it's possible to run gemini on any repository. It might be gitlab collected by borges without rovers. For us it's 1 line of code, so I think better to add it. Description updated.

PR #25, opened 2017-12-22
URLs to the output of report and query commands

smacker:
Fixes https://github.com/src-d/gemini/issues/9

### DIFF
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -44,9 +44,10 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
       .getFirstReferenceCommit
       .getTreeEntries
       .getBlobs
-      .select("blob_id", "repository_id", "path")
+      .select("blob_id", "repository_id", "commit_hash", "path")
       .withColumnRenamed("blob_id", "blob_hash")
       .withColumnRenamed("repository_id", "repo")
+      .withColumnRenamed("commit_hash", "ref_hash")
       .withColumnRenamed("path", "file_path")
 
   def save(files: DataFrame): Unit = {
@@ -142,7 +143,25 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
   }
 }
 
-case class RepoFile(repo: String, file: String, sha: String)
+object URLFormatter {
+  val services = Map(
+    "github.com" -> "https://%s/blob/%s/%s",
+    "bitbucket.org" -> "https://%s/src/%s/%s",
+    "gitlab.com" -> "https://%s/blob/%s/%s"
+  )
+  val default = ("", "repo: %s ref_hash: %s file: %s")
+
+  def format(repo: String, ref_hash: String, file: String): String = {
+    val urlTemplateByRepo = services.find { case (h, _) => repo.startsWith(h) }.getOrElse(default)._2
+    val repoWithoutSuffix = repo.replaceFirst("\\.git$", "")
+
+    urlTemplateByRepo.format(repoWithoutSuffix, ref_hash, file)
+  }
+}
+
+case class RepoFile(repo: String, ref_hash: String, file: String, sha: String) {
+  override def toString(): String = URLFormatter.format(repo, ref_hash, file)
+}
 
 case class DuplicateBlobHash(sha: String, count: Long) {
   override def toString(): String = s"$sha ($count duplicates)"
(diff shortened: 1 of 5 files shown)


smacker:
Yeah. The issue was about github only. But it's better to support other services too. Bitbucket, gitlab, do we have anything else? Maybe retrieval team can help with a list of services? If I have time next week I'll make more general solution here.

dpordomingo:
You may ask @vmarkovtsev ; he demoed something like that for Apolo

vmarkovtsev:
AFAIK we do not have a fixed list, the repos can be from any source. I short-circuited for now: https://github.com/src-d/apollo/blob/master/apollo/query.py#L82

vmarkovtsev:
BTW, I would suggest using https://github.com/HubSpot/jinjava so that we can share our report templates.

dpordomingo:
Thanks for sharing that link!
I agree on the benefit of shareing conclusions of our investigations and I agree on the benefit of using a templating system
But I'm not sure about using the same templates/code. Since we are in an early stage of the development, it could be an overhead trying to maintain shared code.

vmarkovtsev:
Since Apollo's side is Python, I can quickly and painlessly adapt to the changes in templates on Gemini side. So you lead the way and I follow here.

smacker:
I think it's a good idea to use jinja style templates if we start to generate html reports. For now, it's just console output, doesn't worth it.

About bitbucket. To generate url for bb we need commit hash. Right now we don't save it in DB.
Do we really want to change schema because of it?

bzz:
Nice improvement!

@smacker sorry for confusion, but it seems that schema needs to be changed anyway even for github case to include a reference - otherwise, could you help me understand how do we point to https://github.com/src-d/engine/blob/v0.1.x/.dockerignore vs https://github.com/src-d/engine/blob/master/.dockerignore ?

I think it would make sense to address here the 2 most common cases: Github and Bitbucket at the same time, #9 was updated with details on URL format.

And if that is correct,  given ref information adding a BitBucket url support should not require much effort.

smacker:
Yes. The only problem was to decide to we want to add commit in DB or not. I don't feel it's up to me :) I'll proceed with the change.

bzz:
@smacker 

Hm, what do you mean by
>  add commit in DB or not

Sorry if I did not explain it well above, but we do *not want* to have commit in DB (yet).

smacker:
@bzz I'm confused now. You said we need to implement bitbucket and better github support here. But we __do__ need commit hash in DB for it.

bzz:
After more discussion, decided to store and use [reference hash](https://github.com/src-d/engine/blob/master/src/main/scala/tech/sourced/engine/Schema.scala#L31) for URLs construction.

smacker:
@bzz 
Could you please confirm that schema changes are correct (repo, ref_hash, file in this order).

And that `commit_hash` is always the same as `ref_hash`.
From my understanding of https://github.com/src-d/engine/pull/200 it is true.
Apollo also uses this field: https://github.com/src-d/apollo/blob/master/apollo/bags.py#L47
But still, I'm not 100% sure because the source code of engine isn't that straightforward for me.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -142,7 +143,23 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+    "github.com" ->  "https://%s/blob/%s/%s",
+    "bitbucket.org" -> "https://%s/src/%s/%s",
+    "gitlab.com" -> "https://%s/blob/%s/%s"
+  )
+  val default = ("", "repo: %s ref_hash: %s file: %s")
+
+  def format(repo: String, ref_hash: String, file: String): String = {
+    services.find{case (h, _) => repo.startsWith(h)}
+      .getOrElse(default)._2.format(repo.replaceFirst("\\.git$", ""), ref_hash, file)
+  }
```
bzz: Probably a nitpick, but method chaining might assist readability and is kind of customary to have whitespace around `{` like in http://twitter.github.io/effectivescala/#Collections-Style

Also, outer `{}` can be skipped in this case as well

bzz:
> And that commit_hash is always the same as ref_hash.
yes, I think after `.getFirstReferenceCommit()` that's the correct way.

smacker:
@bzz do you mean to add test case about correct `ref_hash`? Maybe I should add one more for `repository` too?

smacker:
readability is improved.
test cases added.

smacker:
Because of no further discussion. Rebased & Merged.

PR #26, opened 2017-12-22
Use engine 0.2

smacker:
fix #22

It should fix the problem with flaky tests.

PR #27, opened 2017-12-26
[WIP] Dockerize + pod.yaml + run on cluster script

smacker:
Based on https://github.com/src-d/gemini/pull/26 (to avoid crashes) and https://github.com/src-d/gemini/pull/24 (to be able to read logs).

This PR:
1) wraps gemini in docker container for easy testing on kubernetes without any proxies.
2) k8s config for pod (right now it uses my image, I need help with how to work with src-d docker repository)
3) convenient script to run gemini on cluster. Like that: `./run_on_cluster hash <path>` or `./run_on_cluster report`

### DIFF
--- a/run_on_cluster
+++ b/run_on_cluster
@@ -21,13 +21,13 @@ DB_HOST=scylladb
 
 case $1 in
 "hash")
-    kubectl exec gemini -- /bin/bash -c "MASTER=spark://p-spark-master:7077 /gemini/hash -h $DB_HOST $2"
+    kubectl exec gemini -- /bin/bash -c "MASTER=spark://p-spark-master:7077 /gemini/hash -h $DB_HOST ${@:2}"
     ;;
 "query")
-    kubectl exec gemini -- /gemini/query -h $DB_HOST $2
+    kubectl exec gemini -- /gemini/query -h $DB_HOST "${@:2}"
     ;;
 "report")
-    kubectl exec gemini -- /gemini/report -h $DB_HOST $2
+    kubectl exec gemini -- /gemini/report -h $DB_HOST "${@:2}"
     ;;
 *)
     # just get into container
--- a/run_on_cluster
+++ b/run_on_cluster
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Convenient script to run gemini on cluster
+
+# N.B:
+# make sure kubectl has the same version as server
+# for MacOs remove client and install like this:
+# brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/a9f7c11c63db20f3746e034849559ef266808cd1/Formula/kubernetes-cli.rb
+
+# create pod if not exists
+if ! kubectl get pods | grep gemini; then
+    kubectl create -f ./k8s/gemini-pod.yaml
+fi
+
+while ! kubectl get pods | grep "gemini.*Running"; do
+    sleep 1
+done
+
+# variables
+DB_HOST=scylladb
+
+case $1 in
+"hash")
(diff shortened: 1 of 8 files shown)


smacker:
More changes:
1) I have updated engine to v0.3.1 because v0.2 doesn't support HDFS.
2) I have removed list repos from hash command because it assumes local file system (open for suggestion how fix it better, if we need it)

I'll keep everything needed for https://github.com/src-d/backlog/issues/1152 in one PR, later if needed I'll split it to separate PRs.

src/main/scala/tech/sourced/gemini/HashSparkApp.scala
```diff
@@ -55,29 +66,15 @@ object HashSparkApp extends App with Logging {
...
-      val filesToWrite = gemini.hash(reposPath, config.limit)
+      val filesToWrite = gemini.hash(reposPath, config.limit, config.format)
       gemini.save(filesToWrite)
       println("Done")
 
     case None =>
       System.exit(2)
   }
-
-  private def listRepositories(path: String, conf: Configuration, limit: Int): Array[Path] = {
```
bzz: 1. if this is removed - it would make sense to cleanup imports as well.
2. we want to keep this, as it's very useful for user feedback

It already uses HDFS FS API that supports local and remote FS, but needs to be improved though to support different formats i.e bucketed .siva files in HDFS.

run_on_cluster
```diff
@@ -0,0 +1,36 @@
...
+    ;;
+"query")
+    kubectl exec gemini -- /gemini/query -h $DB_HOST "${@:2}"
+    ;;
+"report")
+    kubectl exec gemini -- /gemini/report -h $DB_HOST "${@:2}"
+    ;;
+*)
+    # just get into container
+    kubectl exec -it gemini /bin/bash
```
dpordomingo: `run_on_cluster reportt` will open you a bash into the container?

I think the following could be useful:
- `run_on_cluster bash` &rarr; open a shell
- `run_on_cluster help | whatever-typo | nothing` &rarr; show the available commands.

How would you see that?

dpordomingo: By the way, I like the idea of having `wrapper-command <old-comand> <same old parameters>`, so you can run the same thing keeping the api of the `<old-command>`.
LQCTM (Looks Quite :sunglasses: To Me )

smacker: it does exactly that already. Like `./run_on_cluster hash -v -f bare hdfs://path`
But it will set up DB & spark master for you and run it inside a container instead of local.

dpordomingo: Yes, I like it a lot

smacker: ahahaha. I thought you suggested this improvement, not just pointed on it. :-D Sorry.

run_on_cluster
```diff
@@ -0,0 +1,36 @@
...
+while ! kubectl get pods | grep "gemini.*Running"; do
+    sleep 1
+done
+
+# variables
+DB_HOST=scylladb
+
+case $1 in
+"hash")
+    kubectl exec gemini -- /bin/bash -c "MASTER=spark://p-spark-master:7077 /gemini/hash -h $DB_HOST ${@:2}"
```
bzz: Extracting configuration to env var at the top of the script i.e `$SPARK_MASTER` allows further users to re-configure the script more easy

bzz:
@smacker I would suggest renaming `./run_on_cluster` to `./gemini-k8s-cluster`.

And eventually we should make sure it's runnable on minikube \w Scylla.

smacker:
Everything is in master already.

PR #30, opened 2018-01-03
Dockerize + pod.yaml for k8s

smacker:
We need docker file to deploy it on our cluster but also useful for external users because it allows running gemini with 1 command as soon as we publish image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+FROM debian:stretch-slim
+
+ARG APACHE_SPARK_VERSION=2.2.0
+ARG HADOOP_VERSION=2.7
+ARG DUMP_INIT_VERSION=1.2.1
+ENV SPARK_NAME=spark-${APACHE_SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}
+ENV DUMP_INIT_DEB_NAME=dumb-init_${DUMP_INIT_VERSION}_amd64.deb
+
+ENV SPARK_DIR /opt/${SPARK_NAME}
+ENV SPARK_HOME /usr/local/spark
+
+RUN mkdir -p /usr/share/man/man1 && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    openjdk-8-jre-headless  ca-certificates-java curl && \
+    rm -rf /var/lib/apt/* && \
+    curl http://d3kbcqa49mib13.cloudfront.net/${SPARK_NAME}.tgz | \
+    tar xzf - -C /opt && \
+    curl -L -o ${DUMP_INIT_DEB_NAME} \
+    https://github.com/Yelp/dumb-init/releases/download/v${DUMP_INIT_VERSION}/${DUMP_INIT_DEB_NAME} && \
+    dpkg -i dumb-init_*.deb && \
+    rm ${DUMP_INIT_DEB_NAME} && \
+    apt-get clean
+
+RUN ln -s $SPARK_DIR $SPARK_HOME
+
+WORKDIR /gemini
+COPY . /gemini
+
+RUN ./sbt assembly && ./sbt package
+
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+CMD ["bash", "-c", "sleep infinity & wait"]
--- a/k8s/gemini-pod.yaml
+++ b/k8s/gemini-pod.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gemini
+  labels:
+    app: gemini
+spec:
+  containers:
+  - name: gemini
+    image: smacker/gemini
+    imagePullPolicy: Always
+  nodeSelector:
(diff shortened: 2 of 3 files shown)


k8s/gemini-pod.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: gemini
+  labels:
+    app: gemini
+spec:
+  containers:
+  - name: gemini
+    image: smacker/gemini
```
smacker: oh. sorry, we need to publish image somewhere else and change url here.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    openjdk-8-jre-headless  ca-certificates-java curl && \
+    rm -rf /var/lib/apt/* && \
+    curl http://d3kbcqa49mib13.cloudfront.net/${SPARK_NAME}.tgz | \
+    tar xzf - -C /opt && \
+    mkdir -p /root/.sbt/launchers/0.13.13 && \
+    curl -L -o /root/.sbt/launchers/0.13.13/sbt-launch.jar http://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.13/sbt-launch.jar && \
+    apt-get remove -y curl && \
+    apt-get clean
+
+RUN ln -s $SPARK_DIR $SPARK_HOME
```
dpordomingo: We're trying to maintain the Dockerfiles with one single `RUN` command.
Are you splitting this into three different ones for dev purposes? If so, it lgtm

smacker: the purpose is docker layers. Because we don't use CI yet for publishing it allows caching of previous steps locally which speed-up building a lot. But even for CI it will speed up the process if we configure it properly.

dpordomingo: For dev purposes I noticed that speed-up and agree on that for sure. For prod purposes, you'll need to deal with Infra team (but I share your pov)

smacker: we build images in travis no? infra doesn't manage travis.
by default travis doesn't use docker cache, but there is a workaround with `docker pull` http://atodorov.org/blog/2017/08/07/faster-travis-ci-tests-with-docker-cache/
not sure it worth to do in our case though, depends on the size of last layer, need to test.

Btw base image for engine also use multiple runs https://github.com/src-d/jupyter-spark-docker/blob/master/Dockerfile
So as engine itself https://github.com/src-d/engine/blob/master/Dockerfile

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+WORKDIR /gemini
+COPY . /gemini
+
+RUN ./sbt assembly && ./sbt package
+
+CMD exec /bin/bash -c "trap : TERM INT; sleep infinity & wait"
```
bzz: @smacker does not block merge, just a quick question - what do you think about using https://github.com/Yelp/dumb-init here?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN mkdir -p /usr/share/man/man1 && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+    openjdk-8-jre-headless  ca-certificates-java curl && \
+    rm -rf /var/lib/apt/* && \
+    curl http://d3kbcqa49mib13.cloudfront.net/${SPARK_NAME}.tgz | \
+    tar xzf - -C /opt && \
+    mkdir -p /root/.sbt/launchers/0.13.13 && \
+    curl -L -o /root/.sbt/launchers/0.13.13/sbt-launch.jar http://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.13/sbt-launch.jar && \
```
bzz: Could you help to understand, what is the be benefit of hard-coding it here (with version number), instead of letting ./sbt wrapper to take care of it below?

smacker: by the time we run `./sbt` there is no curl anymore. It can't download jar and recommends to download it manually as I do here. I can just keep curl in container then we can remove it if you prefer that.

Dockerfile
smacker: interesting. even trap isn't really necessary here for now. But if we are going to use this image with docker run it makes sense to add dumb-init.

Dockerfile
dpordomingo: If `sbt` needs `curl` to do its work, I'd consider `curl` as a `sbt` dependency and leave it in the container

smacker: `sbt` itself doesn't need curl. the script to install `sbt` needs it. anyway i'll keep`curl` in container.

smacker:
I'm keeping curl in container now and removed hard-coded sbt launcher version. 
Also it uses `dump-init` now.

smacker:
I would prefer to merge now. And change the url in the task with CI.

smacker:
Oh. We actually have merged cluster script already. And it assumes we have k8s config.
I'm merging this PR. URL will be changed in https://github.com/src-d/gemini/issues/33

PR #35, opened 2018-01-04
Convenient script to run gemini on cluster

smacker:
Comments from wip pr have been addressed.

### DIFF
--- a/gemini-k8s-cluster
+++ b/gemini-k8s-cluster
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+# Convenient script to run gemini on cluster
+#
+# N.B:
+# make sure kubectl has the same version as server
+
+# variables
+declare -r DB_HOST='scylladb'
+declare -r SPARK_MASTER='spark://p-spark-master:7077'
+
+# create pod if not exists
+if ! kubectl get pods | grep gemini > /dev/null; then
+  kubectl create -f ./k8s/gemini-pod.yaml || exit 1
+fi
+
+while ! kubectl get pods | grep "gemini.*Running" > /dev/null; do
+  sleep 1
+done
+
+case $1 in
+  hash)
+    kubectl exec gemini -- /bin/bash -c "MASTER=$SPARK_MASTER /gemini/hash -h $DB_HOST ${@:2}"
+    ;;
+  query)
+    kubectl exec gemini -- /gemini/query -h $DB_HOST "${@:2}"
+    ;;
+  report)
+    kubectl exec gemini -- /gemini/report -h $DB_HOST "${@:2}"
+    ;;
+  bash)
+    kubectl exec gemini -it /bin/bash
+    ;;
+  *)
+    cat >&2 << END;
+Available commands:
+hash [options] <path> - hashes given set of Git repositories, either from container's FS or HDFS
+query [options] <path> - finds duplicate file among hashed repositories
+report [options] - shows all duplicated files among hashed repositories
+bash - opens bash inside gemini container on cluster
+END
+    ;;
+esac


smacker:
I have tried. The problem is now it looks different from our other scripts 🤦‍♂️

gemini-k8s-cluster
```diff
@@ -0,0 +1,41 @@
...
+# N.B:
+# make sure kubectl has the same version as server
+
+# variables
+declare -r DB_HOST='scylladb'
+declare -r SPARK_MASTER='spark://p-spark-master:7077'
+
+# create pod if not exists
+if ! kubectl get pods | grep gemini > /dev/null; then
+  kubectl create -f ./k8s/gemini-pod.yaml
```
dpordomingo: should we take care of errors that could occur?

gemini-k8s-cluster
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+while ! kubectl get pods | grep "gemini.*Running" > /dev/null; do
+  sleep 1
```
dpordomingo: `while` loop will run forever if the pod cannot be created, won't it?

gemini-k8s-cluster
```diff
@@ -0,0 +1,41 @@
...
+    kubectl exec gemini -- /gemini/query -h $DB_HOST "${@:2}"
+    ;;
+  report)
+    kubectl exec gemini -- /gemini/report -h $DB_HOST "${@:2}"
+    ;;
+  bash)
+    kubectl exec gemini -it /bin/bash
+    ;;
+  *)
+    echo "Available commands:"
```
dpordomingo: Since the "default" behavior responses to a wrong command, what if the `echo` outputs to `STDERR` as suggested by the Google style guide?

gemini-k8s-cluster
```diff
@@ -0,0 +1,41 @@
...
+  kubectl create -f ./k8s/gemini-pod.yaml
+fi
+
+while ! kubectl get pods | grep "gemini.*Running" > /dev/null; do
+  sleep 1
+done
+
+case $1 in
+  hash)
+    kubectl exec gemini -- /bin/bash -c "MASTER=$SPARK_MASTER /gemini/hash -h $DB_HOST ${@:2}"
```
dpordomingo: If the `80chars` limit is a hard limit, there are some lines that should be modified.

gemini-k8s-cluster
smacker: yes. because creating can take second, minutes or hours. It depends on many factors. You will have to investigate it manually anyway.

gemini-k8s-cluster
smacker: I think in this particular case it's better to exceed the limit.

gemini-k8s-cluster
smacker: fixed

gemini-k8s-cluster
smacker: fixed

gemini-k8s-cluster
dpordomingo: If I understood you when the [`kubectl create`](https://github.com/src-d/gemini/pull/35/files#diff-87240ec88599237cfcde88be7dd27039R14) exits with `0`, there is no feasible way to know whether the pod will be created, or it won't, or it will take forever. If so, I'm ok with the infinite `while` loop.

smacker: yes. `kubectl create` just creates task to create container in kubernetes. If it can't reach k8s for example it will exist with error. But otherwise it will just notify "ok, now we will try to create". And creation itself can fail because of many reasons:
- wrong configuration
- docker hub is unavailable
- no resources on cluster
- broken image
- million other reasons

it's possible to check for some other statuses besides `Success`. Like `CrashLoopBackoff` but I don't want to handle all possible statuses here. They can be different depends on k8s version and you will have to look what went wrong manually anyway.

Issue #33, opened 2018-01-04
Add CI hook to push image to docker hub


smacker:
Let's decide about tags/branch on which it should happen later.

The current proposal is to build images from 2 branches:
- master
- experimental

smacker:
Let's wait for Maximo opinion about this one first: https://github.com/src-d/ci/pull/50

dpordomingo:
I'd vote "yes"

bzz:
Have asked Maximo to review, shall be merged any time soon!

bzz:
src-d/ci#50 is merged now.

smacker:
@bzz could you please add necessary environment variables in travis?

bzz:
@smacker done

bzz:
@smacker thank you for taking care of this - could you also please assign this issue to yourself?

PR #40, opened 2018-01-18
add docker-push to travis

smacker:
Fixes: #33

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ target/
 metastore_db/
 scylla
 derby.log
+
+/Makefile.main
+/.ci
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+# Package configuration
+PROJECT = gemini
+
+# Including ci Makefile
+MAKEFILE = Makefile.main
+CI_REPOSITORY = https://github.com/src-d/ci.git
+CI_FOLDER = .ci
+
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -7,4 +7,7 @@ target/
 
 metastore_db/
 scylla
-derby.log
\ No newline at end of file
+derby.log
+
+/Makefile.main
+/.ci
```
dpordomingo: Unexpected EOF

Makefile
```diff
@@ -0,0 +1,13 @@
+# Package configuration
+PROJECT = gemini
+
+# Including ci Makefile
+MAKEFILE = Makefile.main
+CI_REPOSITORY = https://github.com/src-d/ci.git
+CI_FOLDER = .ci
```
dpordomingo: Why are we using this names and import rule? It seems to be compared from somewhere but [the current recommendations](https://github.com/src-d/ci/blob/master/examples/ci/Makefile)

.gitignore
smacker: how can I reproduce it? travis works just fine

Makefile
smacker: https://github.com/bblfsh/dashboard/blob/master/Makefile#L7
don't see much difference between it and the example. they do the same thing.

Issue #38, opened 2018-01-17
Release v0.0.1


bzz:
Setup release automation if needed and produce a first release.

 - tag
 - release notes
 - GH release
 - CI automation on tag
 - push image to DockerHub (handled separately under #33 )

smacker:
what do you want to upload to GitHub as release? 3 jars?

bzz:
Basically, yes - an archive with jars and sh wrappers that would allow external users to try out gemini + instructions how to do that, given the release archive.

smacker:
@bzz need to add `GITHUB_TOKEN` environment variable to travis settings.

bzz:
@smacker done

smacker:
@bzz do you what to use the same wrappers as we have right now?

bzz:
Yes, I belive same files should be fine by now. 
What do you think?

smacker:
it requires to put everything related to `sbt` in tarball. And dependencies will be downloaded on user machine. Do we want it? Or do we want to have everything in jar?

smacker:
Agreed IRL:
- use current scripts
- put everything that is needed for sbt in tarball
- user will have to have local spark to run `hash` command

PR #41, opened 2018-01-23
Prepare v0.0.1

smacker:
Fixes: https://github.com/src-d/gemini/issues/38

As agreed it uses sbt. I think we better pack all dependencies in jar file later.
The release is actually there already: https://github.com/src-d/gemini/releases
because I was testing travis.


.travis.yml
```diff
@@ -20,12 +20,47 @@ matrix:
 
   - scala: 2.11.2
     env: INTEGRATION_TESTS=true
+    install:
+      - make build
     script:
-     - ./scripts/get_apache_spark.sh "2.2.0" "2.7" || travis_terminate 1
-     - ./scripts/wait-for-it.sh 127.0.0.1:9042 -- echo "Cassandra/ScyllaDB is up"
```
bzz: if not used, let's remove the file from FS as well

.travis.yml
```diff
@@ -20,12 +20,47 @@ matrix:
...
+      - cp sbt gemini-$TRAVIS_TAG/
+      - cp build.sbt gemini-$TRAVIS_TAG/
+      - cp hash gemini-$TRAVIS_TAG/
+      - cp query gemini-$TRAVIS_TAG/
+      - cp report gemini-$TRAVIS_TAG/
+      - cp project/build.properties gemini-$TRAVIS_TAG/project/
+      - cp project/Dependencies.scala gemini-$TRAVIS_TAG/project/
+      - cp project/plugins.sbt gemini-$TRAVIS_TAG/project/
+      - cp -r src/main/resources/* gemini-$TRAVIS_TAG/src/main/resources/
+      - tar -cvzf gemini_$TRAVIS_TAG.tar.gz gemini-$TRAVIS_TAG/
```
bzz: To keep CI logic visible, let's move all these steps to `./scripts/release.sh`

smacker:
comments are addressed! Thank you for the feedback!

Issue #51, opened 2018-03-14
FE: feature extractors


bzz:
This is an umbrella issue for [Feature Extractors (FE)](https://docs.google.com/document/d/1cInzLuz-qygwrTvddGIbzBnzF4nnsAuk_24a6Nhff0k/edit#heading=h.vjihr4gw6m8r) implementation:

 - [x] Define gRPC Service, messages in `src/main/proto/*.proto` https://github.com/src-d/gemini/pull/52
 - [x] Generation scripts: server in `python src/main/python/generated.py`,  #57 
    Python CLI app `src/main/python/main.py` configures the port
 - [x] Generation scripts: client in scala `src/main/scala` #63 
 - [x] Docker file \w feature extractor (used on CI) #73 
 - [x] Implement FE service: call `sourced.ml.extractros` #73 
 - [x] Add missed weight parameter to extractors #79
 - [x] Integration Tests: add CI profile, that for a given UAST checks the response is not failing #81

smacker:
everything is implemented! 🎉 
I'm closing the issue. If we would need any further improvements later, we will address them in separate issues.

PR #57, opened 2018-03-15
FE: Generate grpc server and create python app

smacker:
Part of #51: 

- Generation scripts: server in python src/main/python/generated.py
- Python CLI app for FE at `src/main/python/server.py` configures the port

I put generation script in `src/main/resources/generate_from_proto.sh` for now but we might need a better place (root Makefile?).


bzz:
Looks nice, 2 cosmetic things:

 1. do you think it could be worth to break down this commit to two: generated files, and everything else? This way the history/blame in Git will be very clear.

 2. `src/main/python` is going to be the place for multiple python apps, i.e CLI for community detection will also live there. Knowing that, do you it may be worth using something different from `main.py` for feature extractors?

src/main/python/generate.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
bzz: I suggest to keep this .sh script in `src/main/resources`, as it's follows the standard project layout, and most probably this script will also include generating gRPC/Protobuf for other languages.

bzz:
One more thing - it's totally fine at this point not to have even a basic tests, but I think we should update a developer documentation https://github.com/src-d/gemini#dev  

Adding one more section about gRPC/Protobuf generation, pointing to `src/main/resourcese/generate.sh` and briefly mentioning what is the input and what is the output of this process and when it should be run.

src/main/python/generate.sh
bzz: And it's also good idea to add a small comment here, following https://google.github.io/styleguide/shell.xml#File_Header

smacker:
Thanks @bzz! Those aren't nitpicking at all!

About tests:
I couldn't figure out what I could test on this point (all methods are raising `NotImplementedError` exception for now). I think it's better to add tests later when there will be implementation.

smacker:
I split code into 3 different commits:
- a script for code generation
- generated code
- app

Moved the script to resources directory.

Renamed main.py to server.py (currently we are going to have only one server, later we may move this file to separate folder inside python, but for now I would like to keep it as simple as possible)

smacker:
I added doc in next PR: https://github.com/src-d/gemini/pull/63

bzz:
Took liberty of updating PR description

bzz:
@carlosms on 

> Instead of requirements.txt, put the requirements in the readme

this file is machine-readable and is part of standard tooling for dependency management on Python with `pip install -r requirements.txt`, so no need to duplicate this in human-readable format IMO.

> Add a stdout message when the server starts, something like 'started listening on port xx'.

👍 not sure if that blocks the merge, but even simplest logging for such things is a very nice engineering practice indeed, that IMO is worth adding at from the beginning of the project

smacker:
added logging

PR #63, opened 2018-03-15
FE: Generate scala client

smacker:
Part of #51 - generation scripts: client in scala at `src/main/scala`


build.sbt
```diff
@@ -32,13 +34,16 @@ assemblyJarName in assembly := s"${name.value}-uber.jar"
...
+  case PathList("scalapb", xs @ _*) => MergeStrategy.first
+  case PathList("com", "trueaccord", "lenses", xs @ _*) => MergeStrategy.first
+  case PathList("com", "google", "protobuf", xs @ _*) => MergeStrategy.first
   case x =>
     val oldStrategy = (assemblyMergeStrategy in assembly).value
     oldStrategy(x)
 }
 
 assemblyShadeRules in assembly := Seq(
-  ShadeRule.rename("com.google.common.**" -> "com.google.shadedcommon.@1").inAll,
```
bzz: I strongly suggests we keep this shading, otherwise chances are very high we'll have problems in Spark runtime at cluster mode.

build.sbt
```diff
@@ -32,13 +34,16 @@ assemblyJarName in assembly := s"${name.value}-uber.jar"
 
 assemblyMergeStrategy in assembly := {
   case "META-INF/io.netty.versions.properties" => MergeStrategy.last
+  // engine uses bblfsh client scala, which also uses scalapb but different version
```
bzz: Wonder if there is a reason we could not declare, document and use same version of scalapb as bblfsh client does?

build.sbt
smacker: It breaks build.
- `./sbt package`
- `./sbt assembly`
- `./sbt package` - fails with error expected `com.google.common` got `com.google.shadedcommon`

let me know how to avoid it without removing shading

build.sbt
smacker: 1. bblfsh client-scala uses quite old version. We use last version 0.7.1. It contains bug fixes and other improvements.
2. I believe it's better to update bblfsh client-scala then stick to old version. Temporary we can live with custom merge strategy for proto files (they aren't needed after scala files are generated so here it could be removed strategy as well)
3. Updating version in bblfsh client-scala is a bit tricky because master isn't compiling right now (I'll take a look why today)

smacker:
@bzz I have rebased code on @carlosms branch with tests on spark in distributed mode. And CI passed. So it looks like it works fine without shadowing `"com.google.common.**"`.

src/main/proto/google/protobuf/descriptor.proto
```diff
@@ -0,0 +1,872 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// Author: kenton@google.com (Kenton Varda)
+//  Based on original Protocol Buffers design by
+//  Sanjay Ghemawat, Jeff Dean, and others.
+//
+// The messages in this file describe the definitions found in .proto files.
+// A valid .proto file can be translated directly to a FileDescriptorProto
+// without any other information (e.g. without reading its imports).
```
bzz: Sorry, may be a stupid question, but I must have missed the explanation of the reason why is this file needed?

src/main/proto/service.proto
```diff
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+package tech.sourced.featurext;
```
bzz: Clearly grouping generated code together is convenient for both, humans and machines i.e for excluding from linters/formatters, so why do not we put it in `tech.sourced.featurext.generated`, what do you think?

src/main/resources/generate_from_proto.sh
```diff
@@ -4,8 +4,23 @@
...
+--python_out=src/main/python/pb --grpc_python_out=src/main/python/pb \
+src/main/proto/service.proto \
+src/main/proto/github.com/gogo/protobuf/gogoproto/gogo.proto \
+src/main/proto/gopkg.in/bblfsh/sdk.v1/uast/generated.proto
+
+# Scala client
+
+wget https://github.com/scalapb/ScalaPB/releases/download/v0.7.1/scalapbc-0.7.1.zip
+unzip scalapbc-0.7.1.zip
+scalapbc-0.7.1/bin/scalapbc -Isrc/main/proto \
```
bzz: It's very nice to have self-contained, end to end example, but in case user already have `scalapbc`available on there env, do you think we could download/cleanup it conditional? 

Here is example how we handle some conditions in with [spark-submit](https://github.com/src-d/gemini/blob/master/hash#L23).

src/main/resources/generate_from_proto.sh
```diff
@@ -4,8 +4,23 @@
...
 
-python3 -m grpc_tools.protoc -I../proto \
---python_out=../python/pb --grpc_python_out=../python/pb \
-../proto/service.proto \
-../proto/github.com/gogo/protobuf/gogoproto/gogo.proto \
-../proto/gopkg.in/bblfsh/sdk.v1/uast/generated.proto
+# Python server
+
+pip3 install grpcio==1.10.0 grpcio-tools==1.10.0
+python3 -m grpc_tools.protoc -Isrc/main/proto \
```
bzz: As soon as it's an interactive script meant to be run by developer on .proto updates, do you think we could have few very simple logs with `echo` in here?

I find scripts that provide small bits of the output very useful for debugging, i.e

```
Installing grpcio
Generating Python code from .proto to src/main/python/pb
scalabp binary is not found, installing from YYY
Generating Scala code from .proto to src/main/scala
```

README.md
```diff
@@ -107,6 +107,11 @@ Tests (with embedded Cassandra)
 ./sbt test
 ```
 
+To generate gRPC code for Feature Extractors from `src/main/proto/*.proto` files:
+```
+./src/main/resources/generate_from_proto.sh
```
bzz: 👍 awesome, thank you for taking care about dev docs!

build.sbt
bzz: Let's keep it like that, after it's proven with https://github.com/src-d/gemini/pull/63#issuecomment-374553457 not to cause us troubles with modern version of engine on the cluster mode, after an awesome work done by @carlosms .

As soon as #72 is merged, I guess a rebase is in order.

src/main/proto/google/protobuf/descriptor.proto
smacker: https://scalapb.github.io/faq.html

> I am getting “Import was not found or had errors”
> 

This for a context who it works with scalapb:
> If you are using sbt-protoc and importing protos like scalapb/scalapb.proto, or common protocol buffers like google/protobuf/wrappers.proto:
> 
> Add the following to your build.sbt:
> 
> libraryDependencies += "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf"
>
> This tells sbt-protoc to extract protos from this jar (and all its dependencies, which includes Google’s common protos), and make them available in the include path that is passed to protoc.

This why we need it to put as a file and commit:
> If you are not using sbt (for example, spbc), then you need to make those files available on the file system.

bzz: Awesome, thank you for explaining, makes perfect sence!

Than, could you please: 
 - take care of appending a license information, as in example shared above, inside the same commit
 - add that awesome explanation on the reasons, why this file is needed from the above, to the git commit message on the new line (here is a [good example](https://github.com/Khan/jenkins-jobs/commit/d5e5c243b66b89129b5836a47be155c0adffc0b3))

src/main/resources/generate_from_proto.sh
smacker: I don't think it's a good idea. We need exact version of scalapbc (another version can potentially generate different code). I can write complex logic with checking for scalapbc, then version, then download if needed and so on. But because this script would be run very very rare I don't see advantages to make it so complex to save few kilobytes of traffic for a developer

src/main/proto/service.proto
smacker: yep. The original idea was featurext won't have any non-generated code. But it might. So changed. 👍

src/main/proto/google/protobuf/descriptor.proto
smacker: not sure I did everything right with the LICENCE. Please help me to check correctness.
(PS requirement of putting it the same commit forced me to `push --force`)

src/main/resources/generate_from_proto.sh
smacker: done

src/main/resources/generate_from_proto.sh
bzz: Good point, checking version in .sh does not make sense indeed. 

Let's then at least check if the zip file is already present in FS - the reason is that having an option for this automation to work without network access is useful and a good engineering practice.

src/main/proto/google/protobuf/descriptor.proto
bzz: 👍  https://github.com/src-d/gemini/pull/63/commits/9f456a73fab947c0112fa36875054848a02b4786#diff-9879d6db96fd29134fc802214163b95aR677 looks gorgeous

bzz:
CI passes now, thank you for updating the tests not to use actual license.

smacker:
The conditional download is done. Merging.

PR #65, opened 2018-03-16
Refactoring: rename Report App

bzz:
`ReportSparkApp` -> `ReportApp` as it's a non-spark application


bzz:
Thank you @smacker for prompt review!

Change is quite trivial, so merging if there is no further discussion

PR #66, opened 2018-03-16
Refactoring: move report representation details out of API

bzz:
Based on #65, it moves the details of report representation from API to the report class.


bzz:
Merging if there is no further discussion

PR #72, opened 2018-03-16
Add CI tests for Spark standalone cluster mode

### LINKED ISSUE #64: Integration Tests on CI in Spark Standalone cluster mode
Right now we have a [profile on CI](https://github.com/src-d/gemini/blob/master/.travis.yml#L22) that does Integration Tests with Apache Spark in local mode.

In order to be able to catch more trickier issues i.e \w runtime classpath collision, we need test it also in Apache Spark cluster mode Standalone configuration.

TODOs:
 - add new profile on CI \w `INTEGRATION_TESTS=true`
 - same test scenario, as in local mode, except for
 - add starting Apache Spark Master and Workers [manually](https://spark.apache.org/docs/latest/spark-standalone.html#starting-a-cluster-manually)
 - run Gemini using, using the above with `MASTER="spark://127.0.0.1:7077" ./hash ...`

carlosms:
Fixes #64

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,22 @@ matrix:
           branch: experimental
 
   - scala: 2.11.2
+    env:
+      - INTEGRATION_TESTS=true
+      - MASTER=spark://127.0.0.1:7077
+    install:
+      - make build
+    script:
+      - ./scripts/get_apache_spark.sh "2.2.0" "2.7" || travis_terminate 1
+      - ./scripts/start_apache_spark_cluster.sh "127.0.0.1" "7077" || travis_terminate 1
+      - sleep 10 # wait for it doesn't work, port is avalable but scyllaDB is still down
+      - ./hash src/test/resources/siva || travis_terminate 1
+      - ./query ./LICENSE
+      - ./report
+      - go get ./src/main/go/... || true
+      - go run ./src/main/go/query.go ./LICENSE
+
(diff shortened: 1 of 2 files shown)


bzz:
Very nice! Thank you for taking care of this improvement.

With changes like this the most important question is - how did it affect the CI time?

From what I can see at https://travis-ci.org/src-d/gemini/builds
Before:
```
Ran for 5 min 2 sec
Total time 8 min 45 sec
```

After:
```
Ran for 8 min 59 sec
Total time 17 min 28 sec
```

My guess is that 17min is a cold-start time without s3 cache.

If you could also please report a warm-cache CI time - that would make this PR even more awesome!

One easy way to do that is to add a trivial change and pushing it to a new branch on top of this one in your fork and check your travis CI profile https://travis-ci.org/carlosms/gemini/builds (must be enabled beforehand).

smacker:
There is an easier way to check time. Just restart job:
https://travis-ci.org/src-d/gemini/builds/354373701

```
 Ran for 5 min 59 sec
 Total time 14 min 25 sec
```

bzz:
> There is an easier way to check time

Does it mean that the logs of the previous CI run will be lost?

smacker:
@bzz yep. Actually most probably old logs are still accessible somehow, but I couldn't find how.

PR #73, opened 2018-03-16
FE: Implement FE service

smacker:
Part of https://github.com/src-d/gemini/issues/51

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ matrix:
     script: ./sbt scalastyle
 
   - scala: 2.11.2
+    env: FE_PYTHON_TEST=true
     script:
       - docker build -t srcd/gemini-fe -f FE.Dockerfile .
       - docker run --rm -e PYTHONHASHSEED=0 srcd/gemini-fe pytest -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,12 @@ matrix:
     env: STYLE_CHECK=true
     script: ./sbt scalastyle
 
+  - scala: 2.11.2
+    script:
+      - docker build -t srcd/gemini-fe -f FE.Dockerfile .
+      - docker run --rm -e PYTHONHASHSEED=0 srcd/gemini-fe pytest -v
+      # add here tests using scala client later
+
 before_script:
  - ./scripts/start_docker.sh
 
--- a/src/main/python/server.py
+++ b/src/main/python/server.py
@@ -8,6 +8,7 @@ import logging
 import time
 
 import grpc
+from sourced.ml.extractors import IdentifiersBagExtractor
 
 import pb.service_pb2 as service_pb2
 import pb.service_pb2_grpc as service_pb2_grpc
@@ -16,15 +17,24 @@ _ONE_DAY_IN_SECONDS = 60 * 60 * 24
 
 
 class Service(service_pb2_grpc.FeatureExtractorServicer):
-    pass
+    """Feature Extractor Service"""
+
+    def Identifiers(self, request, context):
+        """Extract identifiers weighted set"""
+
+        id_extractor = IdentifiersBagExtractor(
+            docfreq_threshold=request.docfreqThreshold, split_stem=request.splitStem)
+
+        features = [service_pb2.Feature(name=f[0], weight=f[1])
+                    for f in id_extractor.extract(request.uast)]
+
+        return service_pb2.FeaturesReply(features=features)
 
 
 def serve(port):
     logger = logging.getLogger('feature-extractor')
 
-    server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
-    service_pb2_grpc.add_FeatureExtractorServicer_to_server(Service(), server)
-    server.add_insecure_port('[::]:%d' % port)
+    server = _get_server(port)
     server.start()
     logger.info("server started on port %d" % port)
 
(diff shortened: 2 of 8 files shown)


carlosms:
Should we add the python tests to travis in this PR?

smacker:
@carlosms what do you mean? they are added

carlosms:
My bad, I scanned the PR quickly thought your changes were testing the docker build only.

.travis.yml
```diff
@@ -54,6 +54,12 @@ matrix:
     env: STYLE_CHECK=true
     script: ./sbt scalastyle
 
+  - scala: 2.11.2
+    script:
```
bzz: It's minor, but for convenience and consistency, could you please add `env` section to this build with something descriptive and human-readable, like `FE_PYTHON_TEST=true` ?

It does not mean much yet, but is super-helpful with TravisCI, when a project has many profiles, to quickly identify the purpose of the failing one. 

Like i.e [here](https://travis-ci.org/src-d/gemini/builds/332875688) one would be able to tell that it's only Integration Tests failing for this job.
screen shot 2018-03-21 at 3 05 06 pm

bzz:
Looks awesome, a quick question - how were the fixtures generated for the FE?

src/main/python/server.py
```diff
@@ -16,15 +18,43 @@
...
 
 class Service(service_pb2_grpc.FeatureExtractorServicer):
-    pass
+    """Feature Extractor Service"""
+
+    def Identifiers(self, request, context):
+        """Extract identifiers weighted set"""
+
+        extractor = IdentifiersBagExtractor(
+            docfreq_threshold=request.docfreqThreshold, split_stem=request.splitStem)
```
bzz: From what I can see, all `BagsExtractor`s, of which `IdentifiersBagExtractor` is a [particular kind](https://github.com/src-d/ml/blob/master/sourced/ml/extractors/identifiers.py#L7), have a [`weight` parameter](https://github.com/src-d/ml/blob/master/sourced/ml/extractors/bags_extractor.py#L18) - do you think it should be added on our side as well?

AFAIK it's something that most probably going to be used as part of parameter adjustment as soon as the values are inferred from a labeled data, provided by 🐈 .

src/main/python/server.py
```diff
@@ -16,15 +18,43 @@
...
+        extractor = LiteralsBagExtractor(
+            docfreq_threshold=request.docfreqThreshold)
+
+        return self._create_response(extractor.extract(request.uast))
+
+    def Uast2seq(self, request, context):
+        """Extract uast2seq weighted set"""
+
+        extractor = UastSeqBagExtractor(
+            docfreq_threshold=request.docfreqThreshold)
```
bzz: Same situation with arguments of the base algorithm class [here](https://github.com/src-d/ml/blob/master/sourced/ml/algorithms/uast_ids_to_bag.py#L21) , allthough these guys `token2index` and `token_parser` class name might be of less value right now and could be added later, as soon as they are really needed.

What do you think?

src/main/python/server.py
```diff
@@ -37,11 +67,24 @@ def serve(port):
...
+
+
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Feature Extractor Service.')
     parser.add_argument("--port", type=int, default=9001,
                         help="server listen port")
     args = parser.parse_args()
 
+    # sourced-ml expects PYTHONHASHSEED != random or unset
+    if os.getenv("PYTHONHASHSEED", "random") == "random":
```
bzz: As far as I can see, these are src-d/ml implementation details, relevant to some extractors. 
What do you think - on our side, would it be wise to make them part of the API and pass them as explicit parameters? 

This could help to decouple a library API with it's runtime environment, which is a good practice anyway.

smacker: Nope. This parameter changes how python interpreter works. You can’t change it in runtime. That’s why the check is in __main__.

src/main/python/server.py
smacker: oh. I'm not sure how I missed it in the task about creating proto files. My bad.

Do you mind if I add this parameter in a separate PR? It should be in all extractors and requires changes in proto & scala.

src/main/python/server.py
smacker: I found only 1 implementation for token parser and token index in ml lib. So I decided, for now, we don't need additional parameters there.

smacker:
You can generate a fixture using sdk. The command for it was added here https://github.com/bblfsh/sdk/pull/249

all comments were addressed (by comments). @bzz could you please take another look? I see only 1 problem: weight param. But I strongly prefer to add it in separate PR to avoid conflicts.

src/main/python/server.py
bzz: Reading https://docs.python.org/3.3/using/cmdline.html#envvar-PYTHONHASHSEED helped to understand it better, which would be nice to have somewhere (i.e in that comment above the code)

bzz:
> You can generate a fixture using sdk. The command for it was added here bblfsh/sdk#249

Really, really nice, thank you for linking 👍 ! I only wish it could be documented somewhere, i.e same as code generation from .proto, so next person has to update the tests or something like that - it saves time for a him/her.

> but I strongly prefer to add it in separate PR to avoid conflicts.

sounds good to me

smacker:
I'm not sure about the rightest place to put info about fixtures for now.

I believe I would need fixtures for integration test scala-python (last task in umbrella issue), it might change the location of fixtures or I'll create some script. I'll document how to generate fixtures in that task when figure out the best way to do it.

src/main/python/server.py
smacker: there is a comment already. 2 lines below.

PR #75, opened 2018-03-16
Add YAPF lint and format to makefile

### LINKED ISSUE #62: Add python formatter support
Same as with Scala in #50 - as soon as we have src/main/python - it would be nice to document and add https://github.com/google/yapf support.

carlosms:
Fixes #62.

Commit 3cbbf5b adds 2 commands, `make lint-python` and `make format-python`. Both use https://github.com/google/yapf with the default style.

Commit c2dce37 simply applies the formatting to the existing code.

The lint should be added to the travis CI, but I'm not sure if we want to do it now or later when we add python tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ MAKEFILE = Makefile.main
 CI_REPOSITORY = https://github.com/src-d/ci.git
 CI_FOLDER = .ci
 
+# Python configuration
+YAPF = yapf
+PYTHON_LINT_DIRS = src/main/python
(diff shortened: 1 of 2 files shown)


smacker:
@carlosms how do you want to proceed with this PR? we have 2 options:

- merge it and later create another issue/pr for CI
- wait a bit for https://github.com/src-d/gemini/pull/73 and then rebase, add CI & merge?

carlosms:
I'd say we can merge it now and start using as soon as possible, and see how well YAPF works for us. Up to @bzz I guess.

bzz:
@carlosms why not adding lint to CI as part of this PR?

Issue #58, opened 2018-03-15
Report: connected components


bzz:
Part of the https://github.com/src-d/gemini/issues/54: update Scala implementation of the report to write a Connected Components of similar-files graph.

TODO:
 - query a DB for candidates [graph.py#find_connected_components()](https://github.com/src-d/apollo/blob/master/apollo/graph.py#L90)
 - detect [connected components](https://github.com/src-d/apollo/blob/master/apollo/graph.py#L62)
 - define Parquet schema and write the results on disk

smacker:
Pr is merged.

Issue #60, opened 2018-03-15
Report: community detection cli app


bzz:
Part of https://github.com/src-d/gemini/issues/54: small app to read connected components, do community detection and pretty-print the results.
 
A CLI app in Python, that reads the connected components from Parquet created in #58, does community detection using the library in #59 and prints the output to STDOUT, in a format consistent \w current duplicate output format.

A call to this app should also be included in `./report`, right after a duplicate detection, so for a end user whole process will look like a single application. 

Not mandatory, but might be a good idea to use some kind of simple templating library for formatting the output, so templates can be shared between JVM and Python.

bzz:
As report implementation was included in the release https://github.com/src-d/gemini/releases/tag/v0.0.2 - closing this issue.

Please, feel free to create a new issue for subsequent improvements of the report command!

carlosms:
Here it is: https://github.com/src-d/gemini/issues/110

PR #77, opened 2018-03-21
Report: Connected Components

smacker:
Solves https://github.com/src-d/gemini/issues/58

### DIFF
--- a/src/main/resources/schema.cql
+++ b/src/main/resources/schema.cql
@@ -2,3 +2,4 @@ CREATE KEYSPACE IF NOT EXISTS __KEYSPACE__ WITH replication = {'class': 'SimpleS
 USE __KEYSPACE__;
 CREATE TABLE IF NOT EXISTS __KEYSPACE__.blob_hash_files (sha1 ascii, repo text, commit ascii, path text, PRIMARY KEY (sha1, repo, commit, path));
 CREATE TABLE IF NOT EXISTS __KEYSPACE__.hashtables (sha1 ascii, hashtable tinyint, value blob, PRIMARY KEY (hashtable, value, sha1));
+
--- a/src/main/resources/schema.cql
+++ b/src/main/resources/schema.cql
@@ -1,3 +1,4 @@
 CREATE KEYSPACE IF NOT EXISTS __KEYSPACE__ WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
 USE __KEYSPACE__;
 CREATE TABLE IF NOT EXISTS __KEYSPACE__.blob_hash_files (sha1 ascii, repo text, commit ascii, path text, PRIMARY KEY (sha1, repo, commit, path));
+CREATE TABLE IF NOT EXISTS __KEYSPACE__.hashtables (sha1 ascii, hashtable tinyint, value blob, PRIMARY KEY (hashtable, value, sha1));
--- a/src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
+++ b/src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
@@ -1,0 +1,164 @@
+package tech.sourced.gemini
+
+import java.nio.ByteBuffer
+
+import com.datastax.driver.core.{Session, SimpleStatement}
+import org.slf4j.{Logger => Slf4jLogger}
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+case class FileHash(sha1: String, value: ByteBuffer)
+
+class DBConnectedComponents(
+                             log: Slf4jLogger,
+                             conn: Session,
+                             table: String,
+                             keyspace: String = Gemini.defautKeyspace)
+  extends ConnectedComponents(log) {
+  def getHashtables(): List[Byte] = {
+    val hashtablesCql = s"SELECT DISTINCT hashtable FROM $keyspace.$table"
+
+    conn
+      .execute(new SimpleStatement(hashtablesCql))
+      .asScala
+      .map(_.getByte("hashtable"))
+      .toList
+      .sorted
+  }
+
+  def getHashValues(hashtable: Byte): Iterable[FileHash] = {
+    val cql = s"SELECT sha1, value FROM $keyspace.$table WHERE hashtable=$hashtable"
+
+    conn.execute(new SimpleStatement(cql))
+      .asScala
+      .map(row => FileHash(row.getString("sha1"), row.getBytes("value")))
+  }
+}
+
+abstract class ConnectedComponents(log: Slf4jLogger) {
+  def getHashtables(): List[Byte]
+
+  def getHashValues(hashtable: Byte): Iterable[FileHash]
+
+  /**
+    * Group sha1-values from hashtables to buckets
+    *
+    * sequential sha1s with the same value in hashtable are be grouped to one bucket
+    * identical sha1s from different hashtables have the same element id
+    *
+    * it allows to operate with simple list of ints later instead of big ByteBuffers
+    *
+    * @return
+    */
+  def makeBuckets(): List[List[Int]] = {
+    val (buckets, elementIds) = getHashtables()
+      .foldLeft(List[List[Int]](), mutable.Map[String, Int]()) {
+        (result, hashtable) =>
+
+          var (buckets, elementIds) = result
+          val prevBucketsSize = buckets.size
+          var band: Option[ByteBuffer] = None
+          var bucket = List[Int]()
+
+          getHashValues(hashtable).foreach { case FileHash(sha1, value) =>
+            val elId = elementIds.getOrElseUpdate(sha1, elementIds.size)
+            if (!band.contains(value)) {
+              if (band.isDefined) {
+                buckets = buckets :+ bucket
+                bucket = List[Int]()
+              }
+              band = Some(value)
+            }
+            bucket = bucket :+ elId
+          }
+
+          if (bucket.nonEmpty) {
+            buckets = buckets :+ bucket
+          }
+
+          val bucketsSize = buckets.size - prevBucketsSize
+          log.info(s"Fetched $hashtable, $bucketsSize buckets")
+
+          (buckets, elementIds)
+      }
+
+    log.info(s"Number of buckets: ${buckets.size}")
+    log.info(s"Number of elements: ${elementIds.size}")
+
+    buckets
+  }
+
+  /**
+    * Create hash map: element to list of bucket's indices
+    *
+    * @param buckets List of buckets
+    * @return
+    */
+  def elementsToBuckets(buckets: List[List[Int]]): Map[Int, List[Int]] = {
+    buckets
+      .zipWithIndex
+      .foldLeft(Map[Int, List[Int]]()) { case (elementToBuckets, (bucket, i)) =>
+        bucket.foldLeft(elementToBuckets) { (elementToBuckets, el) =>
+          var list = elementToBuckets.getOrElse(el, List[Int]())
+          list = list :+ i
+          elementToBuckets + (el -> list)
+        }
+      }
+  }
+
+  /**
+    * Iterate over buckets and create list of connected components
+    *
+    * Steps:
+    *  - make the list of unvisited buckets
+    *  - make a map of connected components (groupId -> list of elements)
+    *
+    *  loop:
+    *  - visit last unvisited bucket
+    *  - create a new connected components group and put elements of the bucket there
+    *  - iterate over elements of the bucket
+    *  - if an element appears in another bucket add elements of that bucket to the cc group
+    *    and mark bucket as visited
+    *  - repeat
+    *
+    * @param buckets List of buckets
+    * @param elementToBuckets Map of elements to bucket indices
+    * @return
+    */
+  def findInBuckets(
+                     buckets: List[List[Int]],
+                     elementToBuckets: Map[Int, List[Int]]): Map[Int, Set[Int]] = {
+
+    var connectedComponentsElement = Map[Int, Set[Int]]()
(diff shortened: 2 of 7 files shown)


src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+                              keyspace: String = Gemini.defautKeyspace) extends ConnectedComponents(log) {
+  def getHashtables(): List[Byte] = {
+    val hashtablesCql = s"SELECT DISTINCT hashtable FROM $keyspace.hashtables"
+
+    conn
+      .execute(new SimpleStatement(hashtablesCql))
+      .asScala
+      .map(_.getByte("hashtable"))
+      .toList.
+      sorted
```
bzz:
```
.toList
.sorted
```

src/main/scala/tech/sourced/gemini/ReportApp.scala
```diff
@@ -47,6 +47,9 @@ object ReportApp extends App {
       val gemini = Gemini(null, log)
       gemini.applySchema(cassandra)
 
+      gemini.findConnectedComponents(cassandra)
+      System.exit(2)
```
bzz: Would anything below be executed?

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  def elementsToBuckets(buckets: List[List[Int]]): Map[Int, List[Int]] = {
+    buckets.zipWithIndex.foldLeft(Map[Int, List[Int]]()) { case (elementToBuckets, (bucket, i)) =>
+      bucket.foldLeft(elementToBuckets) { (elementToBuckets, el) =>
+        var list = elementToBuckets.getOrElse(el, List[Int]())
+        list = list :+ i
+        elementToBuckets + (el -> list)
+      }
+    }
```
bzz: Do you think this method can be further improved by refactoring it a little bit, taking into consideration http://twitter.github.io/effectivescala/#Collections-Style ?

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+}
+
+abstract class ConnectedComponents(log: Slf4jLogger) {
+  def getHashtables(): List[Byte]
+
+  def getHashValues(hashtable: Byte): Iterable[FileHash]
+
+  def makeBuckets(): List[List[Int]] = {
+    val (buckets, elementIds) = getHashtables()
+      .foldLeft(List[List[Int]](), collection.mutable.Map[String, Int]()) {
```
bzz: If it's important to have a mutable version here for some reason, here and everywhere else it's better to follow http://twitter.github.io/effectivescala/#Formatting-Imports

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
+package tech.sourced.gemini
+
+import java.nio.ByteBuffer
+
+import com.datastax.driver.core.{Session, SimpleStatement}
+import org.slf4j.{Logger => Slf4jLogger}
+import scala.collection.JavaConverters._
+import scala.collection.mutable.ListBuffer
```
bzz: Not sure why it's important to have mutable collection here, but here and below, it's good idea to follow http://twitter.github.io/effectivescala/#Collections-Use

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+
+case class FileHash(sha1: String, value: ByteBuffer)
+
+class DBConnectedComponents(
+                              log: Slf4jLogger,
+                              conn: Session,
+                              keyspace: String = Gemini.defautKeyspace) extends ConnectedComponents(log) {
```
bzz: Here and below, for the class declarations, its a good idea to check that the style guide is followed http://docs.scala-lang.org/style/declarations.html#classes 

In this case, the preferred way would be:
```scala
class DBConnectedComponents(
    log: Slf4jLogger,
    conn: Session,
    keyspace: String = Gemini.defautKeyspace) 
  extends ConnectedComponents(log) {
}
```

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -122,6 +124,13 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
     ReportExpandedGroup(duplicates)
   }
 
+  def findConnectedComponents(conn: Session): Unit = {
```
bzz: AFAICS, it would be reasonable to have concrete return type of `Map[Int, Set[Int]]` as part of gemini API here.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+    buckets.zipWithIndex.foldLeft(Map[Int, List[Int]]()) { case (elementToBuckets, (bucket, i)) =>
+      bucket.foldLeft(elementToBuckets) { (elementToBuckets, el) =>
+        var list = elementToBuckets.getOrElse(el, List[Int]())
+        list = list :+ i
+        elementToBuckets + (el -> list)
+      }
+    }
+  }
+
+  def findInBuckets(
```
bzz: It may be a good idea to add some [ScalaDoc](http://twitter.github.io/effectivescala/#Formatting-Comments) with very high-level description of the steps of the algorithm, so the reader has some idea of what is being done below.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+      .map(row => FileHash(row.getString("sha1"), row.getBytes("value")))
+  }
+}
+
+abstract class ConnectedComponents(log: Slf4jLogger) {
+  def getHashtables(): List[Byte]
+
+  def getHashValues(hashtable: Byte): Iterable[FileHash]
+
+  def makeBuckets(): List[List[Int]] = {
```
bzz: It may be a good idea to add some [ScalaDoc](http://twitter.github.io/effectivescala/#Formatting-Comments) with very high-level description of the steps of the algorithm, so the reader has some idea of what is being done below.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,125 @@
...
+    conn
+      .execute(new SimpleStatement(hashtablesCql))
+      .asScala
+      .map(_.getByte("hashtable"))
+      .toList.
+      sorted
+  }
+
+  def getHashValues(hashtable: Byte): Iterable[FileHash] = {
+    val cql = s"SELECT sha1, value FROM $keyspace.hashtables WHERE hashtable=$hashtable"
```
bzz: Do you think it would be possible for `hashtables` table, to follow the convention we use for [meta table](https://github.com/src-d/gemini/blob/master/src/main/scala/tech/sourced/gemini/Gemini.scala?utf8=%E2%9C%93#L180)?

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: I moved extends to a new line. But indent is still large, IntelliJ IDEA does it.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: sorry, I don't really get the idea behind meta table. But I added table name as a parameter.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: this is the only place where I use mutable structure. Imo, it fits here. It's possible to rewrite using the immutable map but it would complicate code without any advantages. Usage of a mutable map is scoped to only this function.
It works similar to cache and using mutable map for cache is recommended in https://docs.scala-lang.org/overviews/collections/maps.html

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: the problem is: there is no intermediate result that I could name. It's `for` loop inside `for` loop in a functional style.

src/main/scala/tech/sourced/gemini/ReportApp.scala
smacker: yeah :) that's why it's WIP :)

smacker:
@bzz thanks a lot for the review! most of the comments are addressed others are answered.
From my point of view, there is only one issue with `meta` table. Could you please provide a small example how do you want to use it?

carlosms:
For the record, a problem we found and discussed via chat:
The parquet file created by the code in this PR cannot be read from python. Tested with `parquet-python`, `fastparquet`, and `pyarrow.parquet`. Apparently Avro may be using some unique or non-standard format.

src/main/resources/schema.cql
```diff
@@ -1,3 +1,4 @@
 CREATE KEYSPACE IF NOT EXISTS __KEYSPACE__ WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};
 USE __KEYSPACE__;
 CREATE TABLE IF NOT EXISTS __KEYSPACE__.blob_hash_files (sha1 ascii, repo text, commit ascii, path text, PRIMARY KEY (sha1, repo, commit, path));
+CREATE TABLE IF NOT EXISTS __KEYSPACE__.hashtables (sha1 ascii, hashtable tinyint, value blob, PRIMARY KEY (hashtable, value, sha1));
```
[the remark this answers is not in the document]
smacker: added.

smacker:
Also:
Despite of problem with parquet file, I would better merge it because it implements querying DB. We can address python/scala communication in separate PR when both this and #85 get merged.

bzz:
@carlosms thank you for useful followup! Indeed, after `pip install pyarrow`

```
import pyarrow.parquet as pq
parquet_file = pq.ParquetFile('cc.parquet') # works
table2 = pq.read_table('cc.parquet')            # breaks
```
on [cc.parquet.zip](https://github.com/src-d/gemini/files/1875130/cc.parquet.zip)



@smacker To make serialization format compatible between different languages, we need to make sure that the schema, produced by this tool, conforms the spec https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists

Right now AFAIK it does not, you can see that by https://github.com/apache/parquet-mr/tree/master/parquet-tools#build and then

`java -jar target/parquet-tools-*.jar schema cc.parquet`

which results in 2-level list encoding, instead of 3-level one from the doc above

```
message cc {
  required group elements (LIST) {
    repeated int32 array;
  }
}
```

According to https://issues.apache.org/jira/browse/PARQUET-1122, there is a setting `conf.set("parquet.avro.write-old-list-structure", "false");` that might help.

If you could try to make it generate the right schema for list encoding and then upload an example of .parquet file here - will be happy to assist with further debugging.

smacker:
oh! thanks a lot for `java -jar target/parquet-tools-*.jar schema cc.parquet`. It wasn't obvious there is such command and what it produces according to readme. :-D

With your awesome tips, I'll try to fix compatibility. I'm changing PR to WIP state.

smacker:
PR is updated. Now I'm able to read generated parquet using pyarrow.

smacker:
rebased

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,164 @@
...
+    * identical sha1s from different hashtables have the same element id
+    *
+    * it allows to operate with simple list of ints later instead of big ByteBuffers
+    *
+    * @return
+    */
+  def makeBuckets(): List[List[Int]] = {
+    val (buckets, elementIds) = getHashtables()
+      .foldLeft(List[List[Int]](), mutable.Map[String, Int]()) {
+        (result, hashtable) =>
```
bzz: May be nice to be consistent with style of how function arguments are specified like i.e in https://github.com/smacker/gemini/blob/d96462356ac84906cb30fc1135f44d320701b604/src/main/scala/tech/sourced/gemini/ConnectedComponents.scala#L63

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -0,0 +1,164 @@
...
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+case class FileHash(sha1: String, value: ByteBuffer)
+
+class DBConnectedComponents(
+                             log: Slf4jLogger,
+                             conn: Session,
+                             table: String,
+                             keyspace: String = Gemini.defautKeyspace)
```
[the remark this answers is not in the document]
smacker: most probably you mean https://docs.scala-lang.org/style/declarations.html#classes

but the long indentation is made by IntelliJ IDEA. We might need https://github.com/src-d/gemini/pull/74 to change it.
But for now, as I remember we agreed to format code using default "reformat code" feature of IntelliJ.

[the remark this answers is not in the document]
bzz: This makes my eyes hurt, but let's do as you suggest.

smacker:
Note: this code does what is described in the issue.
But it might or might not be enough for finding communities. If so, improvements will be handled in #60 or in a separate task.

Issue #67, opened 2018-03-16
Query: CPU WMH implementation in Scala 


bzz:
Part of #53 and #55 

For both, query and hashing similar files we need to have a CPU-based [WeightedMinHash](https://blog.sourced.tech/post/minhashcuda/) implementation in JVM, to avoid depending on MinHashCUDA lib and GPU.

Here are few reference implementations of this algorithm:
 - [Go impl](https://github.com/src-d/go-license-detector/blob/d69b5eb70a3f26cb1261ad01fd2f51934f06075a/licensedb/internal/wmh/wmh.go#L64)
 - [Python impl](https://github.com/ekzhu/datasketch/blob/ae9edad7b43ab6eccd86537cc9bfdea11fddb485/datasketch/weighted_minhash.py#L106)
 - [C++ CUDA impl](https://github.com/src-d/minhashcuda/blob/e0aa2758292a4c527a9fea0e25bcd4f4af13b39c/minhashcuda.cc#L564)

It might make sense to research if such library already exists in JVM ecosystem, and if not - it might make sense to implement one in Java as part of Gemini codebase. This way it can be later used from Java, Scala, Clojure, etc eventually, as a standalone library.

Correctness verification is of paramount importance, so we would need to have some tests, and may be a reference data \w hashes of some sets, produced by one of the implementations from above.

carlosms:
Closing, code merged.

PR #78, opened 2018-03-23
Weighted MinHash

carlosms:
_Description edited: now it works, and tests are passing_

Issue #67.

Right now there is a java class, `WeightedMinHash`, based on these two implementations:
* https://github.com/ekzhu/datasketch/blob/master/datasketch/weighted_minhash.py
* https://github.com/src-d/go-license-detector/blob/master/licensedb/internal/wmh/wmh.go

If I'm not mistaken, we'll get a Compressed Sparse Row matrix from the sourced.ml lib. The current implementation takes a dense array, like the datasketch implementation. I thought it would be better to start here, and then modify it to work with sparse arrays or a sparse matrix (maybe using a lib, [like this one](https://github.com/fommil/matrix-toolkits-java/blob/master/src/main/java/no/uib/cipr/matrix/sparse/CompRowMatrix.java)).

Here's what I did to test it. I implemented similar tests to the [MinHashCuda testing](https://github.com/src-d/minhashcuda/blob/master/test.py).

There is a python script in `src/test/resources/weighted-minhash/weighted_minhash_test.py` that uses the datasketch lib to create a bunch of csv files. The scala test `WeightedMinHashSpec` reads those files to initialize the `WeightedMinHash` with the same `rs`, `lnCs`, and `betas`, eliminating any difference in the random initialization of python and java. I didn't add the the csv files to git because there are _a lot_ and they weight over 20MB.

There is a new travis configuration to run only this test, as it a bit different because it depends on python. This travis job takes some time to install all the python dependencies, maybe we could find a way to cache those.


vmarkovtsev:
Sure, I will review the code in the upcoming days!

bzz:
@carlosms really good job! Thank you for taking care.

Sorry for these niggles, but would you also be able to provide a CI time stats before/after? 
That would help to determine if some form of travis caching would be needed as part of this PR as well.

carlosms:
As you can see [in Travis](https://travis-ci.org/src-d/gemini/builds/358419594) (PYTHON_DEPENDENCY_TESTS) it takes about the same time as the integration tests, so the total time is not increased.

However, I know you'll appreciate the new commit 2ec46b5 I have just pushed. It removes the execution of the python minhash code, and instead a new .tar.gz file with the fixtures is stored in the repo. WeightedMinHashSpec is now tested in the main travis task, and it adds no noticeable extra time. To compare the times, take a look at [build 200](https://travis-ci.org/src-d/gemini/builds/358419594) and [build 218](https://travis-ci.org/src-d/gemini/builds/361754477).

smacker:
@carlosms could you please rebase (there are some conflicts) and merge?

PR #80, opened 2018-03-26
Add python yapf lint to travis CI

### LINKED ISSUE #61: Add python linter to CI
As we have `src/main/python`, it would be nice to have something like a https://www.pylint.org/ integrated in CI.

carlosms:
Fixes #61.

Adds a new travis job that runs `make lint-python`, and another commit that fixes current formatting.

The manual python installation is needed because travis.yml can only have one language environment, in our case scala.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,15 @@ matrix:
       - docker run -p 9001:9001 -e PYTHONHASHSEED=0 -d srcd/gemini-fe
       - ./sbt "test-only * -- -n tags.FEIntegration"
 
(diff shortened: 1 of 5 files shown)


smacker:
@bzz 
- there is no cache for apt-get in travis: https://docs.travis-ci.com/user/caching#Things-not-to-cache
- linting happens in separate travis task (mean in parallel) and takes much less time than some other tasks. You can see it in the report for this PR: https://travis-ci.org/src-d/gemini/builds/358506566?utm_source=github_status&utm_medium=notification

Issue #68, opened 2018-03-16
Query: UAST extraction


bzz:
Part of the #53 

Add UAST extraction using https://github.com/bblfsh/client-scala to the query API

PR #82, opened 2018-03-28
Query: UAST extraction

smacker:
Fixes https://github.com/src-d/gemini/issues/68

This PR implements UAST extraction and logs result. That's it for now. Here is almost none of our own code, so it should be tested only by integration tests.


src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -70,19 +73,41 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
     val path = new File(inPath)
     if (path.isDirectory) {
       findDuplicateProjects(path, conn, keyspace)
       //TODO: implement based on Apollo
       //findSimilarProjects(path)
     } else {
+      findSimilarForFile(path, bblfshClient)
       findDuplicateItemForFile(path, conn, keyspace)
       //TODO: implement based on Apollo
       //findSimilarFiles(path)
```
bzz: May be this comments are not relevant any more and could be removed?

src/main/scala/tech/sourced/gemini/QueryApp.scala
```diff
@@ -46,7 +55,8 @@ object QueryApp extends App {
       val gemini = Gemini(null, log)
       gemini.applySchema(cassandra)
 
-      val similar = gemini.query(file, cassandra)
+      val client = BblfshClient.apply(config.bblfshHost, config.bblfshPort)
+      val similar = gemini.query(file, cassandra, client)
```
bzz: Just a suggestion - up to you, but merely for the sake of consistency, as Cassandra client var is called `cassandra`, it may be worth calling bblfsh client `bblfsh` here.

bzz: Seems like this might have been addressed in https://github.com/src-d/gemini/pull/83/commits/6e82586ce94060e7d8e13c5d7110498d9ee26656#diff-768357618a8db6a4da444273c800e8d9R68

smacker: yep. Absolutely agree and addressed in next pr.

smacker:
I removed the unrelevant  comment. Renaming is happening in next pr. Merging.

smacker:
rebased

bzz:
Thank you @smacker ! Looks great, let's wait for @carlosms and :shipit:

Issue #69, opened 2018-03-16
Query: extract Features


bzz:
Part of the #53 

Call Feature Extractors using gRPC client, to get feature from the UASTs.

PR #83, opened 2018-03-28
Query: extract Features

smacker:
Fixes: https://github.com/src-d/gemini/issues/69
Based on #82 

Similar doesn't do much. Just calling service and logging result.


bzz:
It seems that FE calls are part of the query command but only a #81 adds gRPC server start on CI, so a bit confused - why CI passes in here.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -108,6 +118,21 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+    val uast2seqRequest = Uast2seqRequest(uast=Some(uast), docfreqThreshold=5)
+    client.identifiers(idRequest)
+
+    val features = for {
+      idResponse <- client.identifiers(idRequest)
+      litResponse <- client.literals(litRequest)
+      uast2seqResponse <- client.uast2Seq(uast2seqRequest)
+    } yield idResponse.features ++ litResponse.features ++ uast2seqResponse.features
+
+    Await.result(features, Duration(30, SECONDS))
```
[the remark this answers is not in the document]
smacker: it would print `log.error` for now. I decided to keep it as `log` message instead of a crush because duplication functionality is still usable without similarity. But I see other possibilities like make it mandatory or configure by a flag. I'm very welcome to any input about it (imo, the best way would be to discuss and create an issue)

But for current integration tests, this code is never executed because we test with `LICENCE` file. It is a text, not a code. So UAST extractor returns empty UAST.

I don't see for now how to write CI tests for query, I tested it only locally. CI is checking that the code is compilable only. I think writing unit tests for 10 lines of code that doesn't do anything but calls api isn't beneficial. The API client is already covered by unit tests.

For now, my plan is:
- merge "UAST extraction" (locally tested)
- merge "extract Features" (locally tested)
- implement "implement actual query logic" and make `findSimilarForFile` actually return results. Write CI tests to check those results. (not as a part of integration testing but as unit tests with mocked dependencies)
- implement reporting (with unit tests) and implement full integration test.

Unit tests would require meaningful fixtures and without "actual query logic" I can't create such.

smacker:
rebased

bzz:
Looks good to me, sans fixing a CI and a suggestion on bbflsh drivers above.

scripts/start_docker_bblfsh.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# CI convenience script to start a bblfsh server in Docker container.
+# Skips starting server for jobs that do not need it i.e stylecheck
+
+E_BAD_DOCKER=2
+image="bblfsh/bblfshd:v2.4.2"
+
+set -ev
+
+if [[ -z "${STYLE_CHECK}" ]]; then
```
carlosms: Maybe we should find a better way to skip docker images? For instance we will have at some point lint for python, scala, and other possible new tests that need a different env.
Would it work if in travis.yml we moved the `before_script:` inside each matrix entry? I would be more verbose, but easier to spot where we are launching docker.
I'm not sure about the best option, but it's something worth looking into.

scripts/start_docker_bblfsh.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Unable to start Docker ${file_list}" >&2
+    exit "${E_BAD_DOCKER}"
+  fi
+
+  docker exec -it bblfshd bblfshctl driver install --recommended
+else
+  echo "Skip starting Docker image ${image}"
+fi
+
+docker ps
```
smola: I understand this is not strictly related to this PR, since you were already doing this previously with `scripts/start_docker_db.sh`, but you might you docker-compose for this.

It can be easily pre-installed in Travis ([see here](https://docs.travis-ci.com/user/docker/#Using-Docker-Compose) and it's a single file with more concise syntax to define Docker dependencies. You can choose to start all of them with `up` or specific `up bblfsh`.

cc @bzz

smacker: yes. We are going to have 3 docker images (db, bblfsh, feature extraction) in dependencies for gemini. It makes perfect sense to use docker-compose in the feature. (actually, I already do locally)
Make it should be part of new release preparation issue?

scripts/start_docker_bblfsh.sh
smacker: current check is copy-past for a script we already use. But I agree is need something better. Maybe also use docker-compose as proposed by smola. Out of scope for this PR though.

@carlosms if you have good ideas for improvement travis, please create an issue with a description because I didn't really think about it yet.

smacker:
@bzz Cache for bblfsh drivers is added:

```
28.27s$ ./scripts/start_docker_bblfsh.sh
if [[ -z "${STYLE_CHECK}" ]]; then
  echo "Starting Docker image ${image}"
  mkdir -p $HOME/bblfsh-drivers
  docker run -v $HOME/bblfsh-drivers:/var/lib/bblfshd --name bblfshd --privileged -p 9432:9432 -d "${image}"
  if [[ "$?" -ne 0 ]]; then
    echo "Unable to start Docker ${file_list}" >&2
    exit "${E_BAD_DOCKER}"
  fi
  docker exec -it bblfshd bblfshctl driver install --recommended
else
  echo "Skip starting Docker image ${image}"
fi
Starting Docker image bblfsh/bblfshd:v2.4.2
Unable to find image 'bblfsh/bblfshd:v2.4.2' locally
v2.4.2: Pulling from bblfsh/bblfshd
Status: Downloaded newer image for bblfsh/bblfshd:v2.4.2
bc65653742e13799eb4932c37b1fa37b398b20a8738a0a7e5cd543904b3ba6c3
Installing ruby language driver from "docker://bblfsh/ruby-driver:latest"... Error
Error, driver already installed: ruby (image reference: bblfsh/ruby-driver:latest)
Installing go language driver from "docker://bblfsh/go-driver:latest"... Error
Error, driver already installed: go (image reference: bblfsh/go-driver:latest)
Installing java language driver from "docker://bblfsh/java-driver:latest"... Error
Error, driver already installed: java (image reference: bblfsh/java-driver:latest)
Installing python language driver from "docker://bblfsh/python-driver:latest"... Error
Error, driver already installed: python (image reference: bblfsh/python-driver:latest)
Installing javascript language driver from "docker://bblfsh/javascript-driver:latest"... Error
Error, driver already installed: javascript (image reference: bblfsh/javascript-driver:latest)
Installing php language driver from "docker://bblfsh/php-driver:latest"... Error
Error, driver already installed: php (image reference: bblfsh/php-driver:latest)
```

PR #85, opened 2018-04-03
Report: community detection library

### LINKED ISSUE #59: Report: community detection library
Part of the https://github.com/src-d/gemini/issues/54: create an internal library for community detection.

Small internal library in Python (a function or class) for community detection using IGraph, based on the algorithm from [graph.py#detect_communities()](https://github.com/src-d/apollo/blob/master/apollo/graph.py#L191).

Should include a very basic tests (but not really testing IGraph implementation itself).

carlosms:
Fixes #59.

There are 2 commits:
- 52ca345 is the relevant one
- 75b7f2b moves the current python code for feature-extractor to a new subfolder

The code basically consists of the apollo code in [graph.py#detect_communities()](https://github.com/src-d/apollo/blob/master/apollo/graph.py#L191) with these differences:
- Instead of a file, the connected components input is given as method arguments
- Instead of writing a file, the output is returned
- There is no option to use spark
- The code has been stripped of all non-essential dependencies

There is a test that uses numpy `.npz` files as input, and as output to compare. These npz files can be regenerated with `generator.py`, but do not need to be created every time the test is run.
The source data is the `ccs.asdf` file. This file was created with the `apollo cc` command using the data from https://github.com/src-d/backlog/issues/1222.

TODO:
- [x] Add new tests to travis.yml
- [ ] ~~Add tests for more (all?) the `edges`, `algorithm`, and `algorithm_params`~~ Maybe later on. The code is ready to run any algorithm, but for now the CLI will only use the default, which is already tested.
- [x] Consider if log should be removed => left with debug level

Question: do we want to support all the `edges`, `algorithm`, and `algorithm_params`? If not, should it be a subset, or a unique default?


src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,141 @@
+from collections import defaultdict
+from itertools import chain
+import logging
+from igraph import Graph
```
smacker: empty line is missed https://www.python.org/dev/peps/pep-0008/#imports

src/main/python/community-detector/fixtures/requirements.txt
```diff
@@ -0,0 +1,6 @@
+numpy==1.14.2
```
smacker: why do you need to specify this deps?
sourced-ml already has them: https://github.com/src-d/ml/blob/master/requirements.txt

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,141 @@
...
+    :param buckets_matrix: scipy.sparse.csr_matrix with the buckets
+    :param edges: The method to generate the graph's edges:
+        - linear: linear and fast, but may not fit some of the CD algorithms, or all to all within a bucket
+        - quadratic: slow, but surely fits all the algorithms.
+    :param algorithm: The community detection algorithm to apply.
+    :param algorithm_params: Parameters for the algorithm (**kwargs, JSON format).
+    :return: JSON with data, indptr
+    """
+
+    log = logging.getLogger("community-detector")
```
smacker: From TODO in PR description:
> Consider if log should be removed

I would keep logging. I see you use info level. The default level is warning, so they won't pollute output. But logs will be useful for debugging.

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def detect_communities(cc,
+                       buckets_matrix,
+                       edges="linear",
+                       algorithm="walktrap",
+                       algorithm_params={}):
+    """
+    Runs Community Detection analysis on the given connected components.
+    :param cc: numpy.ndarray with the connected components
```
smacker: according to code looks like it doesn't have to be `numpy.ndarray`. But if I'm wrong, please specify shape.

src/main/python/community-detector/test_community_detector.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def build_csr_matrix(input):
+    return csr_matrix(
+        (input['id_to_buckets_data'], input['id_to_buckets_indices'],
+         input['id_to_buckets_indptr']),
+        shape=input['id_to_buckets_shape'])
+
+
+class TestServer(unittest.TestCase):
```
smacker: server?

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,141 @@
...
+    :param buckets_matrix: scipy.sparse.csr_matrix with the buckets
+    :param edges: The method to generate the graph's edges:
+        - linear: linear and fast, but may not fit some of the CD algorithms, or all to all within a bucket
+        - quadratic: slow, but surely fits all the algorithms.
```
smacker: do you think we should check `edges` param in code as it is in doc? Currently code accepts `linear` or `1` as linear and everything else as quadratic

bzz:
@carlosms nice work! A bit confused by: 
 - WIP in Issue title - this usually communicates "do not look here, it's not ready yet". Is that the case?
 - un-checked TODO items. What is the status of those?

One more thing - could you also please make sure that the code follows the style guild, like https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments etc ?

Also, in case if some major bodies of code come from another project - it's a good idea to keep an attribution by putting a link in comments, etc.

As soon as it's addressed - please ping for another pass.

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,141 @@
...
+
+def detect_communities(cc,
+                       buckets_matrix,
+                       edges="linear",
+                       algorithm="walktrap",
+                       algorithm_params={}):
+    """
+    Runs Community Detection analysis on the given connected components.
+    :param cc: numpy.ndarray with the connected components
+    :param buckets_matrix: scipy.sparse.csr_matrix with the buckets
```
smacker: could you please document relationship of cc & buckets?

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,154 @@
...
+
+    ccs = defaultdict(list)
+
+    for i, c in enumerate(cc):
+        ccs[c].append(i)
+
+    buckindices = buckets_matrix.indices
+    buckindptr = buckets_matrix.indptr
+    total_nvertices = buckets_matrix.shape[0]
+    linear = edges in ("linear", "1")
```
smacker: `edges == "linear"`. according to the check above it can't be 1.

smacker:
about:
> Question: do we want to support all the edges, algorithm, and algorithm_params? If not, should it be a subset, or a unique default?

I would keep it possible pass different params, but wrote tests only for the subset we really need for our purposes. (also document it)

@bzz what do you think?

carlosms:
@bzz said:

> @carlosms nice work! A bit confused by:
> 
> WIP in Issue title - this usually communicates "do not look here, it's not ready yet". Is that the case?

I wanted to say that the PR is not finished, but I wanted to get some feedback and also had some questions on how to continue. I'll stop using the WIP in the title and just write a proper sentence in the description from now on...

> un-checked TODO items. What is the status of those?

Those are the things I'd like to do before merging. The unchecked items are not done yet.

> One more thing - could you also please make sure that the code follows the style guild, like https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments etc ?

My bad, I was not aware that we are following the Google Python style for docs. and just used the same comment style as Apollo (which I believe uses sphinx style). This is now changed in 6a19934.

> 
> Also, in case if some major bodies of code come from another project - it's a good idea to keep an attribution by putting a link in comments, etc.

Very true. Added in 6a19934.

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,154 @@
...
+
+def detect_communities(cc,
+                       buckets_matrix,
+                       edges="linear",
+                       algorithm="walktrap",
+                       algorithm_params={}):
+    """Runs Community Detection analysis on the given connected components.
+
+    Based largely on the Apollo detect_communities() code from
+    https://github.com/src-d/apollo/blob/master/apollo/graph.py
```
bzz: Grr, GH is not showing reviews from commits in the PR https://github.com/src-d/gemini/commit/6a1993492f49f47e83c211d2d7bf6d90260b3444#r28488078 :/

src/main/python/community-detector/fixtures/requirements.txt
```diff
@@ -0,0 +1,5 @@
+python-igraph==0.7.1.post6
+scipy==1.0.1
+cassandra_driver >= 3.12.0, <4.0
```
bzz: Do we use Cassandra somewhere in fixture generation?

carlosms: because apollo is not available in pip, it is cloned in ./apollo. Instead of installing apollo/requirements.txt, the minimum dependencies to call apollo are included here.

Without cassandra driver, it breaks with:
```
Traceback (most recent call last):
  File "generator.py", line 5, in <module>
    from apollo.graph import ConnectedComponentsModel, CommunitiesModel, detect_communities
  File "./apollo/apollo/graph.py", line 17, in <module>
    from apollo.cassandra_utils import get_db, patch_tables, BatchedHashResolver, Session
  File "./apollo/apollo/cassandra_utils.py", line 9, in <module>
    from cassandra.cluster import Cluster, Session, NoHostAvailable
ModuleNotFoundError: No module named 'cassandra'
```

bzz:
@carlosms 

> Question: do we want to support all the edges, algorithm, and algorithm_params? 

Not sure if I understood it right, but by [looking at the code](https://github.com/src-d/gemini/pull/85/files#diff-933a11e98dd114abd208aa301a278bc8R9) it seems like this library already supports and exposes on API-level exactly the same set of algorithms and params as Apollo, which is very good. 

If the question is weather those should be exposed to the Gemini CLI - I guess that should be take care of in subsequent PRs https://github.com/src-d/gemini/issues/60 and the answer is that for now, only a hard-coded defaults should be used, same ones as apollo does by default.

Does it answer your question?

src/main/python/community-detector/community_detector.py
```diff
@@ -0,0 +1,154 @@
...
+    for i, community in enumerate(communities):
+        data[pos:pos + len(community)] = community
+        pos += len(community)
+        indptr[i + 1] = pos
+
+    return {"data": data, "indptr": indptr}
+
+
+class CommunityDetector:
+    def __init__(self, algorithm, config):
```
bzz: As discussed, it would be nice to have a references to origins of the code, coming from other projects i.e https://github.com/src-d/apollo/blob/0a6f445a799231f749a7d70610fb53e06ab02763/apollo/graph.py#L267 here.

src/main/python/community-detector/fixtures/requirements.txt
bzz: This is only for tests generation, so it's fine.

But this makes me think that may be all of `ConnectedComponentsModel`, `CommunitiesModel ` and even `CommunityDetector` are a candidates to be proposed refactored out of `graph.py` (to minimize dependency footprint) in the future.

carlosms:
@bzz thanks for the feedback, it should all be addressed now.

> Does it answer your question?

Yes thank you, I wanted to confirm if the current code is OK. In the design doc. we had "1 algorithm, no pySpark", so I wanted to check if you wanted to force the lib to use 1 algorithm only. Limiting it later in the CLI as you said makes sense.

carlosms:
@bzz, @smacker, please take another look when you have time.

src/main/python/community-detector/fixtures/generator.py
```diff
@@ -0,0 +1,73 @@
...
+    ])
+
+    try:
+        handler = args.handler
+    except AttributeError:
+
+        def print_usage(_):
+            parser.print_usage()
+
+        handler = print_usage
```
smacker: simpler:
```python
handler = lambda _: parser.print_usage()
```

src/main/python/community-detector/test_community_detector.py
```diff
@@ -0,0 +1,41 @@
...
+from scipy.sparse import csr_matrix
+
+from community_detector import detect_communities
+
+dirname = os.path.dirname(__file__)
+
+
+def build_csr_matrix(input):
+    return csr_matrix(
+        (input['id_to_buckets_data'], input['id_to_buckets_indices'],
```
smacker: I try to avoid using `input` as a variable name in python. Because it's built-in function: https://docs.python.org/3/library/functions.html#input

src/main/python/community-detector/fixtures/generator.py
carlosms: Much nicer, thank you. This was a copy&paste of pieces from apollo.

src/main/python/community-detector/test_community_detector.py
carlosms: Super useful heads up, thank you!

bzz:
Great job, @carlosms ! Should be good to merge now

PR #86, opened 2018-04-06
scalafmt lint

carlosms:
Partially replaces #74, see https://github.com/src-d/gemini/pull/74#issuecomment-379299888.

Adds a new `make lint-scala` command that uses scalafmt. Travis uses this new command to replace Scalastyle.

PR #87, opened 2018-04-06
scalafmt format

carlosms:
Partially replaces #74, see https://github.com/src-d/gemini/pull/74#issuecomment-379299888.

This branch was created from #86, the relevant commit is just bd39545.


bzz:
Looks great, let's finish and merge the linter first and then get back and :shipit:

smola:
Before merging scalafmt and perform any reformatting, we should discuss if we would like to do for `default` or `intellij` styles. I have also seen that the following might need to be added to enforce line length:

```
# disable case arrow alignment, it breaks line length
align.tokens = []
```

It would be desirable if:
- Our `.scalafmt.conf` does not conflict with previous `scalastyle-config.xml` (actually, both might be kept, since scalastyle checks for stuff scalafmt doesn't.
- It produces minimal reformatting on existing code (this might not be possible, but...).
- We can push this company-wide for Scala projects.

PR #89, opened 2018-04-10
Implement similarity query logic

### LINKED ISSUE #70: Query: implement actual query logic
Part of the #53 

Add [query.py#query()](https://github.com/src-d/apollo/blob/57e52394783d73e38cf1f862afc0166724991fd5/apollo/query.py#L14) logic to Gemini Scala API:

 - no need to handle `args.id` case
 - need to agree with ML team on a way to read WMH params like `htnum` and `band_size` that must be the same, as used for hashing
 - use a DB from a Apollo hash run
 - use vocabulary in .asdf (OrderedDocumentFrequencies), built by Apollo hash \w `--docfreq`

smacker:
fixes #70 

- Precise logic isn't implemented, according to Vadim it doesn't help much and we most probably don't need it.
- Documentation for utilities is only inside source code because they are temporary until we implement hash on our side.
- Test uses fixtures to avoid any external dependencies and failing due to external changes (it will be covered by integration test later)
- Current implementation 100% compatible with apollo and uses apollo results for tests
  - apollo commit: 6b370b5f34ba9e31cf3310e70a2eff35dd978faa
  - source{d} ml commit: 585015146d24522796f189250b6bb08cc77a8566
 - Default log level was changed to print warnings

Task for integration test: https://github.com/src-d/gemini/issues/93


src/main/java/tech/sourced/gemini/WeightedMinHash.java
```diff
@@ -77,7 +77,7 @@ public WeightedMinHash(int dim, int sampleSize, long seed) {
         }
     }
 
-    protected WeightedMinHash(int dim, int sampleSize, double[][] rs, double[][] lnCs, double[][] betas) {
+    public WeightedMinHash(int dim, int sampleSize, double[][] rs, double[][] lnCs, double[][] betas) {
```
bzz: A nitpick, but in cases like this it's usually better to widen the access modifier to a default one, which is a _package-private_ and should work in this case, by just deleting `protected`.

src/main/python/asdf_to_json/main.py
```diff
@@ -0,0 +1,77 @@
...
+import sys
+import argparse
+
+from modelforge.model import Model
+from modelforge.models import register_model
+from sourced.ml.models import OrderedDocumentFrequencies
+
+
+@register_model
+class WeightedMinHashParameters(Model):
```
bzz: For maintainability and future contributors - if a substantial body of code comes from another project, it's better to past a permalink here i.e in comments. 

It's easy to get such link i.e like in https://github.com/src-d/gemini/commit/6a1993492f49f47e83c211d2d7bf6d90260b3444#r28488078

src/main/java/tech/sourced/gemini/WeightedMinHash.java
smacker: default one should be `WeightedMinHash(int dim)` or `WeightedMinHash(int dim, int sampleSize, long seed)`. I use this one temporary while we don't have our own hash.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -81,39 +83,112 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+
+    extractUAST(file, bblfshClient) match {
+      case Some(uast: Node) => {
+        log.debug(s"uast received: ${uast.toString}")
+        val features = extractFeatures(feClient, uast)
+        log.debug(s"features: ${features.toString}")
+        val wmh = hashFile(features, new File(docFreqFile), new File(paramsFile))
+
+        // FIXME try to rewrite later
+        var bands = Array.empty[Array[Byte]]
```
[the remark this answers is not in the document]
smacker: yes. First I want to try to rewrite it in a simpler way. Need more research how and is it possible in scala or not at all. If I wouldn't be able to simplify it, I'm thinking to extract it to a method, document and also write test.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -81,39 +83,112 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+                          conn: Session,
+                          bblfshClient: BblfshClient,
+                          feClient: FeatureExtractor,
+                          docFreqFile: String,
+                          paramsFile: String,
+                          htnum: Int,
+                          bandSize: Int): Option[Array[String]] = {
+
+    extractUAST(file, bblfshClient) match {
+      case Some(uast: Node) => {
```
bzz: If only one brach is taken, it's usually just

```scala
extractUAST() foreach { node =>
  extractFeatures(node)
  hashFile()
  generateBands()
  lookupSimilar()
}
```

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -81,39 +83,112 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
-    }
+    try {
+      val resp = client.parse(file.getName, new String(byteArray))
+      if (resp.errors.nonEmpty) {
+        val errors = resp.errors.mkString(",")
+        log.error(s"bblfsh errors: ${errors}")
+      }
 
-    resp.uast
+      resp.uast
```
bzz: Not related to the changes, but I think it's safer to return `None` in case there were any `resp.errors` present as well.

smacker: I return uast based on how bblfsh works.

LA team can't confirm which drivers work like this right now, but the original idea was:
- bblfsh can parse partly incorrect code (for example if there is error only in one function, this function would be skipped but the rest of code will be parsed correctly)
- such errors go to `resp.errors`

Though I agree it's arguable that we need partial parsing in gemini or not.
Please confirm that we shouldn't return uast in case of any error taken in account info above.

src/main/scala/tech/sourced/gemini/QueryApp.scala
```diff
@@ -38,11 +46,23 @@ object QueryApp extends App {
...
+    opt[String]("doc-freq-file")
+      .action((x, c) => c.copy(docFreqFile = x))
+      .text("path to file with feature frequencies")
+    opt[String]("params-file")
+      .action((x, c) => c.copy(paramsFile = x))
+      .text("path to file with feature frequencies")
+    opt[Int]("hashtables-num")
+      .action((x, c) => c.copy(hashtablesNum = x))
+      .text("number of hashtables should match docFreq file")
+    opt[Int]("band-size")
```
bzz: I think for these two, `hashtables-num` and `band-size` - let's not to expose though a CLI and just use a hard-coded constants with same defaults as in Apollo (hash).

We going to adjust the defaults later on, based on ML research.

smacker: this params right now depend on vocab size. While we don't have our own hash we need to be able to pass the same params as apollo calculates. But later yes, we will be able to remove them from CLI.

src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
```diff
@@ -107,6 +149,96 @@ class CassandraSparkSpec extends FlatSpec
...
+        val iter = featuresMap.map { case(key, value) =>
+          Feature(name = key, weight = value.toInt)
+        }
+        iter.toSeq
+      }
+      case _ => throw new Exception("incorrect json")
+    }
+  }
+
+  "Query for similar files" should "return 1 file" in {
```
bzz: The test looks awesome to me:
 - it uses mocks for FE and Bblfsh
 - it loads to DB from .json fixtures for two hashed files, one of which is src/test/resources/models.py
 - FE extractor mock always return features, extracted for this file
 - and then it checks that 1 similar file is returns

👍

src/main/scala/tech/sourced/gemini/Gemini.scala
bzz: Thank you for the followup!

> LA team can't confirm which drivers work like this right now, but the original idea was:

That is the thing.. the goal for us though is just to be consistent between hash/query. 

So any of the ways would do for now, it's just that in my previous experience working with bblfsh, it was safer and more reproducible to fail on any error.

smacker: yes. I couldn't find code that checks for errors in src-d ml. So to be consistent with hash I also return partial UAST. We can change it later when we implement hash on our side.

thanks for heads up though!

PR #92, opened 2018-04-12
Query output

### LINKED ISSUE #71: Query: include similar files in CLI output
Part of the #53

Update `./query` CLI output to include not only duplicates, but also a similar files.

Same as in report #60, it may (or may not yet) make sense to start using some kind of [templating](https://github.com/HubSpot/jinjava).

smacker:
Fixes #71 
Based on #89

- rename of `defaultTable` to be compatible with apollo
- cli output similar to what we have right now
- added a modified file to test resources and also test with it

### DIFF
--- a/src/main/python/asdf_to_json/main.py
+++ b/src/main/python/asdf_to_json/main.py
@@ -66,18 +66,8 @@ models = {
     }
 }
 
-if __name__ == '__main__':
-    parser = argparse.ArgumentParser()
-    parser.add_argument("asdf", help="path to .asdf file")
-    parser.add_argument(
-        "-m",
-        "--model",
-        required=True,
-        help="src-d.ml model",
-        choices=["docfreq", "params"])
 
-    args = parser.parse_args()
-
+def main(args):
     dirname = os.path.dirname(args.asdf)
     filename = os.path.basename(args.asdf)
     name = os.path.splitext(filename)[0]
@@ -89,3 +79,17 @@ if __name__ == '__main__':
 
     with open(json_path, 'w') as outfile:
         json.dump(j, outfile)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument("asdf", help="path to .asdf file")
+    parser.add_argument(
+        "-m",
+        "--model",
+        required=True,
+        help="src-d.ml model",
+        choices=["docfreq", "params"])
+
+    args = parser.parse_args()
+    main(args)
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -406,7 +406,7 @@ object Gemini {
   val defaultCassandraPort: Int = 9042
   val defaultSchemaFile: String = "src/main/resources/schema.cql"
   val defautKeyspace: String = "hashes"
-  val defaultTable: String = "blob_hash_files"
+  val defaultTable: String = "meta"
   val defaultBblfshHost: String = "127.0.0.1"
   val defaultBblfshPort: Int = 9432
   val defaultFeHost: String = "127.0.0.1"
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -89,25 +89,36 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
             docFreqPath: String = "",
             paramsFilePath: String = "", // TODO remove when we implement hash
             htNum: Int = 0,
-            bandSize: Int = 0): Iterable[RepoFile] = {
+            bandSize: Int = 0): (Iterable[RepoFile], Iterable[RepoFile]) = {
     val path = new File(inPath)
     if (path.isDirectory) {
-      findDuplicateProjects(path, conn, keyspace)
       //TODO: implement based on Apollo
-      //findSimilarProjects(path)
+      (findDuplicateProjects(path, conn, keyspace), findSimilarProjects(path))
     } else {
+      val duplicates = findDuplicateItemForFile(path, conn, keyspace)
+      var similarShas = Array[String]()
+
       if (docFreqPath != "" && paramsFilePath != "") {
-        // TODO implement correct CLI output
-        findSimilarForFile(path, conn, bblfshClient, feClient, docFreqPath, paramsFilePath, htNum, bandSize) match {
-          case Some(similarSha1s: Array[String]) if similarSha1s.nonEmpty =>
-            log.info(s"Similar sha1s: ${similarSha1s.mkString(",")}")
-          case _ => log.info("No similar sha1s")
-        }
+        similarShas =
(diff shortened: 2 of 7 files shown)


src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
```diff
@@ -172,15 +172,34 @@ class CassandraSparkSpec extends FlatSpec
       .start()
 
     val gemini = Gemini(sparkSession, logger, SIMILARITIES)
-
-    val file = new File("src/test/resources/models.py")
     val channel = ManagedChannelBuilder.forAddress("localhost", server.getPort).usePlaintext(true).build()
     val bblfshStub = BblfshClient("localhost", server.getPort)
     val feStub = FeatureExtractorGrpc.stub(channel)
 
-    val similar =
```
bzz: Do you think it could be worth to refactor this single test to two separate ones now?

i.e something like
 - `"Query for duplicate files" should "return 1 file"`
 - `"Query for similar files" should "return 1 file"`

Asking, as I find useful having each test to test only a single case/behaviour.

src/main/python/asdf_to_json/main.py
```diff
@@ -0,0 +1,91 @@
...
+        'class': WeightedMinHashParameters,
+        'transform': lambda x: {
+            'rs': x.rs.tolist(),
+            'ln_cs': x.ln_cs.tolist(),
+            'betas': x.betas.tolist(),
+        },
+    }
+}
+
+if __name__ == '__main__':
```
smola: You could wrap code here in a `main` function and call it from `if __name__ == '__main__:`.

This would make it easier to, eventually, test the main function itself.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -81,39 +83,120 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+                          file: File,
+                          conn: Session,
+                          bblfshClient: BblfshClient,
+                          feClient: FeatureExtractor,
+                          docFreqFile: String,
+                          paramsFile: String,
+                          htnum: Int,
+                          bandSize: Int): Option[Array[String]] = {
+
+    val features = extractUAST(file, bblfshClient).map { uast =>
```
smola: You could split this code in smaller private functions. That would help readability.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -81,39 +83,120 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+      val resp = client.parse(file.getName, new String(byteArray))
+      if (resp.errors.nonEmpty) {
+        val errors = resp.errors.mkString(",")
+        log.error(s"bblfsh errors: ${errors}")
+      }
 
-    resp.uast
+      resp.uast
+    } catch {
+      case e: Exception => {
```
smola: [`NonFatal(e)`](http://www.scala-lang.org/api/current/scala/util/control/NonFatal$.html) would be more appropriate than `e: Exception`.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -122,13 +205,90 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+        log.error(s"feature extractor error: ${e.toString}")
+        Iterable[Feature]()
+      }
+    }
+  }
+
+  def parseDocFreq(file: File): (Int, Map[String, Double], List[String]) = {
+    val docFreqByteArray = Files.readAllBytes(file.toPath)
+    val docFreqJson = JSON.parseFull(new String(docFreqByteArray))
```
smola: You might want to encapsulate this JSON parsing logic in something like:

```scala
def mustParseJSON[T: ClassTag](input: String): T = {
  JSON.parseFull(input) match {
    case Some(res: T) => res
    case Some(_) => throw new Exception("incorrect json")
    case None => throw new Exception("can't parse json")
  }
}
```

Then just:

```scala
val docFreqJson = mustParseJSON[Map[_, _]](new String(docFreqByteArray))
```

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -122,13 +205,90 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+      case Some(_) => throw new Exception("incorrect json")
+      case None => throw new Exception("can't parse json")
+    }
+  }
+
+  def parseParams(file: File): Map[String, List[List[Double]]] = {
+    val paramsByteArray = Files.readAllBytes(file.toPath)
+    val paramsJson = JSON.parseFull(new String(paramsByteArray))
+    paramsJson match {
+      case Some(m: Map[_, _]) => m.asInstanceOf[Map[String, List[List[Double]]]]
```
smola: The above JSON parsing function could then be reused here.

src/main/scala/tech/sourced/gemini/Gemini.scala
smola: My code would need some changes to actually be equivalent to your code with the `.asInstanceOf[Map[String, List[List[Double]]]]` stuff. But the idea would still be the same.

src/main/python/asdf_to_json/main.py
smacker: this is a temporary script while we don't have our own `hash`. It isn't going to have any tests or anything but changed.

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: sorry, do you suggest to extract
```
    val features = extractUAST(file, bblfshClient).map { uast =>
      log.debug(s"uast received: ${uast.toString}")

      val result = extractFeatures(feClient, uast)
      log.debug(s"features: ${result.toString}")

      result
    }
```
and the rest in separate functions?

I agree about the rest, but the code above is actually a one-liner but with `log.debug`. I don't see advantages of extraction for it.

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: I kept `mustParseJSON` as simple as you suggested because it's also temporary code until we have our own hashing.

smacker:
Thank you @smola for the review! All scala comments are actually not for this PR but for https://github.com/src-d/gemini/pull/89

I made changes there. If you have time please check it there.

smacker:
@smola a bit more info why we need json at all now:
- hashing is implemented in apollo only for now
- apollo uses seed of current time by default
- apollo produces params for WMH and docFreq files in asdf format
- there is no implementation for reader of adsf in any language beside python
- json is the simplest format for debugging
- both files will be removed in the future:
  - hashing & query in gemini will use static seed, it will be possible to calculate params for WMH just based on that seed
  - docFreq will be saved to parquet or database

smola:
@smacker Thanks for the explanation!

Issue #88, opened 2018-04-10
FE: add graphlet extractor


bzz:
Right now, a [number of feature extractors](https://github.com/src-d/gemini/blob/master/src/main/proto/service.proto#L19) is already supported and used by default in Gemini.

This task is about: 
 - adding one more extractor from sourced-ml - [`GraphletBagExtractor`](https://github.com/src-d/ml/blob/d06eac30fe4fe35524a20d7265445f4f281c4413/sourced/ml/extractors/graphlets.py#L7)and 
 - change [Gemini defaults to use 3 extractors](https://github.com/src-d/gemini/blob/c7016f512d9aff0ca3e508157c14ca2c76209294/src/main/scala/tech/sourced/gemini/Gemini.scala#L119): `Graphlet`, `Identifier` and `Literals`

PR #96, opened 2018-04-12
FE: add graphlet extractor

smacker:
Fixes: #88


bzz:
@smacker could you please also sign-off a2781da? Otherwise DCO check is un-happy

Issue #76, opened 2018-03-21
Release: include LICENSE file in the release artifacts


bzz:
As discussed in https://github.com/src-d/gemini/pull/63#discussion_r176015006 , this project is governed by GPL3 but include few (well identified, .proto) files under BSD-like licenses, thus we need to make sure we keep copyright notice/disclamer and redistribute it as a part of the final released artefacts.

To do so, we just need to:
 - make sure LICENSE file does not miss any details of all third-party licenses from external resources like `src/main/proto`, etc
 - update the release automation https://github.com/src-d/gemini/blob/master/scripts/release.sh to include LICESE file to the [distributed archive](https://github.com/src-d/gemini/releases).

PR #97, opened 2018-04-13
Add LICENCE file to tarball

smacker:
Fix: #76

There are only 2 external proto files, both of them are mentioned in LICENCE file already.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -680,4 +680,7 @@ BSD-style licenses
 The following components are provided under a BSD-style license. See project link for details.
 
 (BSD 3 Clause) protocol buffers (https://github.com/google/protobuf)
+  src/main/proto/google/protobuf/descriptor.proto
+
 (BSD 2 Clause) gogoprotobuf (https://github.com/gogo/protobuf)
+  src/main/proto/github.com/gogo/protobuf/gogoproto/gogo.proto
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -16,4 +16,5 @@ cp project/build.properties gemini-$VERSION/project/
 cp project/Dependencies.scala gemini-$VERSION/project/
 cp project/plugins.sbt gemini-$VERSION/project/
 cp -r src/main/resources/* gemini-$VERSION/src/main/resources/
+cp LICENCE gemini-$VERSION/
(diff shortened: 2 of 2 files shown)


bzz:
Nice, thank you for taking care!

The goal of this issue to enable us to make a proper source/binary releases from a legal standpoint. 

While we are at it - could you please make sure all code in the repository is:
 - either written by us (or coming from one of the src-d projects and thus is GPL3)
 - or have a license header with clear mention of it's original license + mention in LICENSE?

i.e it's not very clear about https://github.com/src-d/gemini/blob/master/src/test/resources/models_changed.py - in cases like this, to be safer, we should add a comment at the top of the file, stating it's license and where is it coming from. 

One more thing - let's add paths to files inside the codebase, that are covered by those external licenses to the https://github.com/src-d/gemini/blob/master/LICENSE#L682

```
(BSD 3 Clause) protocol buffers (https://github.com/google/protobuf)
  src/main/proto/google/protobuf/descriptor.proto

(BSD 2 Clause) gogoprotobuf (https://github.com/gogo/protobuf)
  src/main/proto/github.com/gogo/protobuf/gogoproto/gogo.proto
```

Thanks again!

smacker:
I don't really know how this legal stuff works. Maybe we should have a meeting about it with the team so you would explain us?

About py files, they aren't part of the release, are they?

bzz:
>  Maybe we should have a meeting about it with the team so you would explain us?

Sure, will be happy to share. Feel free to send cal invite, or we could also add it to the next retrospective agenda.

> About py files, they aren't part of the release, are they?

that's the thing - as we do not have very strict policy for publishing release artifacts defined AND we already distribute a .zip file with full source code of the project, as a part of our our [releases](https://github.com/src-d/gemini/releases/tag/v0.0.1) - they might be. 

In my experience with legal -  it's way easier to stay safe and just preemptively keep track of things, thus the request for changes.

smacker:
> the next retrospective agenda
sounds good

bzz:
Should be ready to merged after feedback from https://github.com/src-d/gemini/pull/97#issuecomment-381856684 is addressed

smacker:
rebased on master, there is no more `model.py` and `model_modified.py` files in the master.
I'm moving PR in CR again. Let me know if we need to do anything else here.

Issue #94, opened 2018-04-12
Release: update release artifacts


bzz:
Right now, Gemini has 3 use cases:

 1. local, for developer on pre-configured environment, using shell scripts
 2. local, for a first-time user, https://github.com/src-d/gemini/issues/84#issuecomment-380751027
 3. k8s \w Apache Spark cluster, Dockerfiles: Gemini \w Spark, Feature Extractors

Current release artifacts are only .tar file for 1 and a Docker file.

This issue is about changing the release process to accommodate recent changes on file-level similarity so:

  - add .sh for starting a feature extractor process on local machine
  - in `./report`, check if Python is available
 - make sure Dockerfile for FE re-uses the shell scripts (as much as possible)
 - add publishing Docker container for FE to the release process
   (this will enable testing Gemini on 3rd use-case)

this way, a new release should accommodate all 1 and 3 case from above.

PR #98, opened 2018-04-13
update release artifacts

smacker:
Fix: #94 

- Fixed gemini docker file
- Copy python source code in tarball
- Bash script to run features-extractor (don't reuse it in docker file. Both are very-very simple and there are like 2 similar commands so unification would require much more code than just keep it as is) 
- Build image & upload image for features-extractor
- Check for python in `report`
- Update sourced-ml to 0.4.5 (0.4.4 contains a strange version of spark and it gets re-downloaded for every run)
- Add  [coursier](https://github.com/coursier/coursier) for faster dependency download

We will need a bit more work after `report` is ready (most probably install python in gemini image, its dependencies and maybe update some scripts). I left it out of this PR for now.


feature_extractor
```diff
@@ -0,0 +1,16 @@
...
+# Convenience script to run features extraction service locally
+
+service_dir="src/main/python/feature-extractor"
+
+hash python3 >/dev/null 2>&1 || { echo "Please install python3" >&2; exit 1; }
+
+if [ -z "$VIRTUAL_ENV" ]; then
+    echo "Looks like you don't use virtual environment for python"
+    echo "Please active it or create a new one"
+    echo "Move information is here: https://virtualenv.pypa.io/en/stable/"
```
carlosms: Small typos:
"Please **activate** it or create a new one"
"**More** information here: https://virtualenv.pypa.io/en/stable/"

smacker:
it's weird that pip couldn't resolve grpcio correctly, constraints are correct...
Anyway, I just set the same version as in bblfsh and now it's ok.

Typos have been fixed.

Further updates will be done after #103 is merged.

smacker:
Python & py deps for report added to Dockerfile. Should be good to go now.

bzz:
```
 Ran for 12 min 5 sec
 Total time 37 min 3 sec
```

🤔

bzz:
Sorry, I do not think this change have to do with test times.

Looks great to me! Let's merge

Issue #93, opened 2018-04-12
Update query integration test


smacker:
Part of the #53

We have one that does hash and then looks for duplicates. We need to update it to test similarity too.

I suggest such a plan:
- Amend current test dataset to include similar file(s)
- Populate DB table `hashtables` with fixture created using apollo with that dataset
- Use docFreq&params from apollo also
- Run bblfsh server & fe server
- Check query output and make sure both duplicates and similar files appear in the output

PR #100, opened 2018-04-17
Query: integration test

smacker:
Fixes: #93 

Use gemini `./hash` for meta data.
Loads hashtables from a dump.
Checks query returns results for both identical & similar files.

Unit test was updated to reuse the same data.


.travis.yml
```diff
@@ -57,13 +57,29 @@ matrix:
       - MASTER=spark://127.0.0.1:7077
     install:
       - make build
+      - 
     script:
+      # start dependencies
       - ./scripts/get_apache_spark.sh "2.2.0" "2.7" || travis_terminate 1
       - ./scripts/start_apache_spark_cluster.sh "127.0.0.1" "7077" || travis_terminate 1
+      # python will be installed after https://github.com/src-d/gemini/issues/60
+      # and after merge https://github.com/src-d/gemini/pull/98 we can use local fe runner
```
bzz: If this PR gets merged first - shall TODOs be added to #60 and #98 to take care of this?

src/test/resources/consumer.go
```diff
@@ -0,0 +1,132 @@
+package borges
```
bzz: As discussed in https://github.com/src-d/gemini/pull/97#issuecomment-381856684 - in my experience, it's safer to be conservative and develop a habit of adding a comment with original license and, if possible, an origin of source code files, if copied from other projects.

.travis.yml
```diff
@@ -57,13 +57,29 @@ matrix:
...
+      - docker build -t srcd/gemini-fe -f FE.Dockerfile .
+      - ./scripts/start_docker_fe.sh
       - sleep 10 # wait for it doesn't work, port is avalable but scyllaDB is still down
       - ./hash src/test/resources/siva || travis_terminate 1
+      # query without similarity
       - ./query ./src/test/resources/LICENSE
+      # query with similarity
+      - docker cp ./src/test/resources/hashetabes_dump.cql db:/hashetabes_dump.cql
+      - docker exec -it db cqlsh -e "SOURCE '/hashetabes_dump.cql'"
+      - ./query --doc-freq-file=src/test/resources/docfreq.json --params-file=src/test/resources/params.json --hashtables-num=9 --band-size=13 ./src/test/resources/consumer.go | tee query_result.txt
```
bzz: `# query with similarity`

.travis.yml
```diff
@@ -57,13 +57,29 @@ matrix:
...
       - ./scripts/start_apache_spark_cluster.sh "127.0.0.1" "7077" || travis_terminate 1
+      # python will be installed after https://github.com/src-d/gemini/issues/60
+      # and after merge https://github.com/src-d/gemini/pull/98 we can use local fe runner
+      - docker build -t srcd/gemini-fe -f FE.Dockerfile .
+      - ./scripts/start_docker_fe.sh
       - sleep 10 # wait for it doesn't work, port is avalable but scyllaDB is still down
       - ./hash src/test/resources/siva || travis_terminate 1
+      # query without similarity
       - ./query ./src/test/resources/LICENSE
+      # query with similarity
```
bzz: `#Load hashtables from apollo DB dump`

.travis.yml
smacker: idk which one is going to be merged first. This PR contains lines that should be replaced later. That's why the comment is here.

src/test/resources/consumer.go
smacker: than it wouldn't be identical

.travis.yml
smacker: it's obvious from the command that it loads dump into DB, no matter actually where the fixture comes from.
I added comments to split groups of commands by what they do, not rewrote the command in English.

bzz:
@smacker shall we merge, after the typo in filename is fixed?

smacker:
fixed the typo, rebased on master. Will merge after CI is green.

PR #90, opened 2018-04-12
Call community detector lib with parquet data

carlosms:
Partially implements #60.

This PR writes extra data in `report`, and adds a new `report.py` command that reads parquet files and calls the community detector python lib.

```bash
$ ./report --keyspace apollo -o src/main/python/community-detector/parquets/
$ python src/main/python/community-detector/report.py src/main/python/community-detector/parquets
```

Next steps, to be done in future PRs:
- save the output data as a parquet file.
- make scala report app call the python command, wait for it to finish, and read the parquet data.
- use this data (possible calling the DB to get extra info) in the scala report output.

Notes:
If you look at `report.py` it would make sense to write a single parquet file with the columns `element_id, cc, buckets`. But as I put in a TODO, the community detector lib actually uses internally `cc->element-ids`. So the current `cc.parquet` could be read and used skipping `build_id_to_cc`. Left as a TODO to avoid changing too much until we have all the parts working together.

PR #103, opened 2018-04-19
Report: use community detector lib, print similar files info

carlosms:
Part of #60, this is a continuation of #90.

In `report.py` the community detection output is written to a new parquet file.
`ReportApp.scala` calls this python command and reads the new parquet file. With the communities of IDs it goes to the DB to fetch info about each file and prints it.

There is a new `scripts/install_python.sh` in order to run the report command in Travis CI.

For this PR the output is pretty simple:
```
No duplicates found.

2 similar files:
	https://github.com/whatupdave/pil/blob/f829a9c2c8d41733c3363e61d8ad9ebdae8072a1/PIL/ImageStat.py
	https://github.com/python-pillow/Pillow/blob/c013188594ba44fcd42e1adca3e4744e3983c2a0/src/PIL/ImageStat.py

2 similar files:
	https://github.com/python-pillow/Pillow/blob/c013188594ba44fcd42e1adca3e4744e3983c2a0/Tests/test_file_fitsstub.py
	https://github.com/python-pillow/Pillow/blob/c013188594ba44fcd42e1adca3e4744e3983c2a0/Tests/test_file_hdf5stub.py
```

### DIFF
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -513,11 +513,13 @@ object Gemini {
     // Transform communities of element IDs to communities of sha1s
     // Equivalent to apollo graph.py BatchedCommunityResolver._gen_hashes
     // https://github.com/src-d/apollo/blob/f51c5a92c24cbedd54b9b30bab02f03e51fd27b3/apollo/graph.py#L295
-    val communitiesSha1 = communities.map {
-      case (communityId, community) => community.collect {
-        case id if (idToSha1.contains(id)) => idToSha1(id)
+    val communitiesSha1 = communities
+      .map { case (_, community) =>
+        community
+          .filter(idToSha1.contains)
+          .map(idToSha1)
       }
-    }.filter(_.size > 1)
+      .filter(_.size > 1)
 
 
     communitiesSha1.map(sha1s => {
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -339,20 +339,28 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
       }
   }
 
+  def reportCommunities(conn: Session,
+                        communities: List[(Int, List[Int])],
+                        elementIds: Map[String, Int]): Iterable[Iterable[RepoFile]] = {
+    getCommunities(conn, keyspace, communities, elementIds)
+  }
+
   /**
     * Return connected components from DB hashtables
     *
     * @param conn Database connections
-    * @return Tuple with:
-    *         - Map of connected components groupId to list of elements
+    * @return - Map of connected components groupId to list of elements
     *         - Map of element ids to list of bucket indices
+    *         - Map of element to ID
     */
-  def findConnectedComponents(conn: Session): (Map[Int, Set[Int]], Map[Int, List[Int]]) = {
+  def findConnectedComponents(conn: Session): (Map[Int, Set[Int]], Map[Int, List[Int]], Map[String, Int]) = {
     val cc = new DBConnectedComponents(log, conn, "hashtables", keyspace)
-    val buckets = cc.makeBuckets()
+    val (buckets, elementIds) = cc.makeBuckets()
     val elsToBuckets = cc.elementsToBuckets(buckets)
 
-    (cc.findInBuckets(buckets, elsToBuckets), elsToBuckets)
+    val result = cc.findInBuckets(buckets, elsToBuckets)
+
+    (result, elsToBuckets, elementIds)
   }
 
   def applySchema(session: Session): Unit = {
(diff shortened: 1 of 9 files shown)


src/main/scala/tech/sourced/gemini/ReportApp.scala
```diff
@@ -69,13 +74,31 @@ object ReportApp extends App {
...
+
+        cassandra.close()
+        cluster.close()
 
-      print(report)
+        print(report)
+        println()
+        printCommunities(reportCommunities)
+      } else {
+        log.error(s"Failed to execute '${pythonCmd}'")
```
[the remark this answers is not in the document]
carlosms: It is not straightforward to capture both the command exit code and the output. But as the code is right now any python output to stdout/stderr is forwarded to the report app stdout.

So if something fails you will see the problem, and then the exact command that was used. Like this:
```
Traceback (most recent call last):
  File "src/main/python/community-detector/report.py", line 3, in <module>
    import numpy
ModuleNotFoundError: No module named 'numpy'
ERROR 16:27:01 tech.sourced.gemini.ReportApp$ (ReportApp.scala:95) - Failed to execute 'python3 src/main/python/community-detector/report.py src/main/python/community-detector/parquets/'
```

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -494,5 +502,36 @@ object Gemini {
...
+    val idToSha1 = for ((elem, id) <- elementIds) yield (id, elem.split("@")(1))
+
+    // Transform communities of element IDs to communities of sha1s
+    // Equivalent to apollo graph.py BatchedCommunityResolver._gen_hashes
+    // https://github.com/src-d/apollo/blob/f51c5a92c24cbedd54b9b30bab02f03e51fd27b3/apollo/graph.py#L295
+    val communitiesSha1 = communities.map {
+      case (communityId, community) => community.collect {
+        case id if (idToSha1.contains(id)) => idToSha1(id)
+      }
+    }.filter(_.size > 1)
```
bzz: It took me a while to parse nested patten matching.

Would this be equivalent of

```scala
    val communitiesSha1 = communities
      .map { case (_, community) =>
        community
          .filter(idToSha1.contains)
          .map(idToSha1)
      }
      .filter(_.size > 1)
```

?

src/main/scala/tech/sourced/gemini/ReportApp.scala
```diff
@@ -166,4 +201,32 @@ object ReportApp extends App {
...
+    val reader = AvroParquetReader.builder[GenericRecord](parquetFilePath).build()
+
+    var result = mutable.Map[Int, List[Int]]()
+
+    var nextRecord: GenericRecord = null
+
+    while ( {
+      nextRecord = reader.read
+      nextRecord
+    } != null) {
```
bzz: Wow, this looks like written in C.

Could you find any other options to re-write this iteration construct?

carlosms: I tried to in c24eab1, other suggestions are welcome.

src/main/scala/tech/sourced/gemini/ReportApp.scala
```diff
@@ -69,13 +74,31 @@ object ReportApp extends App {
         case `groupByMode` => ReportExpandedGroup(gemini.reportCassandraGroupBy(cassandra))
       }
 
-      val (connectedComponents, elsToBuckets) = gemini.findConnectedComponents(cassandra)
+      val (connectedComponents, elsToBuckets, elementIds) = gemini.findConnectedComponents(cassandra)
       saveConnectedComponents(connectedComponents, elsToBuckets, config.ccDirPath)
 
-      cassandra.close()
-      cluster.close()
+      val pythonCmd = s"python3 src/main/python/community-detector/report.py ${config.ccDirPath}"
```
bzz: I suggest a very simple refactoring of this method, so it's logic is more transparent and maintainable in a long run:

```scala
// cassandra & gemini setup

val duplicateReport = ...
val similarityReport = ... // what is called `reportCommunities` now

print(duplicateReport)
printCommunities(similarityReport)
// cassandra teardown
```

src/main/scala/tech/sourced/gemini/ReportApp.scala
```diff
@@ -166,4 +200,25 @@ object ReportApp extends App {
...
   }
 
+  def readCommunities(dirPath: String): List[(Int, List[Int])] = {
+    val parquetFilePath = new Path(s"$dirPath/communities.parquet")
+    val reader = AvroParquetReader.builder[GenericRecord](parquetFilePath).build()
+
+    Iterator
+      .continually(reader.read)
+      .takeWhile(_ != null)
+      .map(record => {
```
bzz: Very minor, but usually in such cases just

```scala
.map { record => 
  // do stuff
}
.toList
```

is enough.

smacker: reader doesn't have `map`. I saw this pattern with `Iterator.continually.takeWhile` in engine and jgit a lot. So I think we can keep it as is.

bzz: :+1: there is just extra `()` that could be avoided here

PR #104, opened 2018-04-19
fix dependencies + assembly deps

smacker:
Multiple (but related to each other) changes:

- split 1 fat jar into 2 jars:
  - one with all external dependencies
  - another one with gemini itself
- update `./hash` script to support deps jar
- `target` directory must be managed, otherwise, `assembly` would try to include old build results in a produced jar. Set it as unmanaged only in the release tarball
- remove MergeStrategy rules for scalapb and protobuf. After we added bblfsh client, the old client from engine gets evicted and there are no conflicts.
- add shading for `com.google.common` and `com.google.protobuf`, because spark also contains this packages but older versions.
- add rename for `com.trueaccord.scalapb`. Bblfsh client from engine got evicted, but engine itself have references for `com.trueaccord.scalapb`. In newer version of scalapb the package was renamed. (but fully compatible with previous one)
- update engine to the last version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN ln -s $SPARK_DIR $SPARK_HOME
 
 COPY . /gemini
 
-RUN ./sbt assembly && ./sbt package
+RUN ./sbt assemblyPackageDependency && ./sbt assembly && ./sbt package
 
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 CMD ["bash", "-c", "sleep infinity & wait"]
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ $(MAKEFILE):
 -include $(MAKEFILE)
 
 build:
+	$(SBT) assemblyPackageDependency
 	$(SBT) assembly
 	$(SBT) package
 
--- a/build.sbt
+++ b/build.sbt
@@ -6,8 +6,7 @@ scalaVersion := "2.11.11"
 version := "0.0.2-SNAPSHOT"
 
 name := "gemini"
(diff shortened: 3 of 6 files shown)


bzz:
Let's rebased on master, after #100 is ✅

smacker:
@bzz rebased

bzz:
@smaker, could you please help me understand - from well the items in PR description, what are the "convenience" changes and what are the fixes for actual compile-time issues?

It might be something obvious, but we might have missed the clear problem description that these changes address, and that seems relevant to  https://github.com/src-d/gemini/pull/105

smacker:
> split 1 fat jar into 2 jars
> update ./hash script to support deps jar

Improve build time for `assembly` (4s instead on 2 min for each rebuild)

> target directory must be managed, otherwise, assembly would try to include old build results in a produced jar. Set it as unmanaged only in the release tarball

it's a bug fix. Currently rebuild of assembly after any changes doesn't work without `./sbt clean`

> remove MergeStrategy rules for scalapb and protobuf. After we added bblfsh client, the old client from engine gets evicted and there are no conflicts.

just remove unnecessary code

> add shading for com.google.common and com.google.protobuf, because spark also contains this packages but older versions.
> add rename for com.trueaccord.scalapb. Bblfsh client from engine got evicted, but engine itself have references for com.trueaccord.scalapb. In newer version of scalapb the package was renamed. (but fully compatible with previous one)

fix runtime errors in spark. (I believe the description of each point is clear enough)

It isn't related to `./sbt test`. You can't create fat jar during `./sbt test`. I have found only 2 works around. Build fat jar and include it as unmanaged during tests and use subprojects for tests (as described in that issue)

PR #105, opened 2018-04-20
[WIP] Hash: generate docFreq

smacker:
WIP


smacker:
Just including deps jar in test doesn't help much:
```
Uncaught exception when running tech.sourced.featurext.ClientSpec: java.lang.AbstractMethodError: gopkg.in.bblfsh.sdk.v1.uast.generated.Node$.parseFrom(Lshaded/com/google/protobuf/CodedInputStream;)Lscalapb/GeneratedMessage;
```
I'll try with additional sub project.

project/Dependencies.scala
```diff
@@ -8,7 +8,8 @@ object Dependencies {
   lazy val spark = "org.apache.spark" %% "spark-core" % "2.2.0"
   lazy val fixNetty = "io.netty" % "netty-all" % "4.1.17.Final"
   lazy val fixNewerHadoopClient = "org.apache.hadoop" % "hadoop-client" % "2.7.2"
-  lazy val engine = "tech.sourced" % "engine" % "0.5.7" classifier "slim"
+  lazy val engine = ("tech.sourced" % "engine" % "0.5.7" classifier "slim")
+    .exclude("com.google.shadedcommon", "util")
```
bzz: Could you please help to understand, why is this exclusion needed?

In general, any library can have things shaded that it needs (of particular version, etc) - and it's implementation detail.  Shading would guarantee class names are uniq and only used inside the library itself (so there should be no collisions).

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -55,16 +55,67 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+      .extractUASTs
+      .select("repository_id", "path", "blob_id", "uast")
+      .filter(_.getAs[Seq[Array[Byte]]]("uast").nonEmpty)
+      .withColumn("document", expr("CONCAT(repository_id, '//', path, '@', blob_id)"))
+      .select("document", "uast")
+  }
+
+  def sparkFeatures(uasts: DataFrame): RDD[SparkFeature] = {
+    var feConfig = SparkFEClient.getConfig(session)
+    val rows = uasts.flatMap { row =>
```
bzz: As apache spark iteration is followed by collection iteration here, adding a suffix to var names would communicate it clearer i.e `uastsDf` and `rowsDf` would contrast to `uastArr` and `features`.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -55,16 +55,67 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
...
+      .classifyLanguages
+      .filter('lang.isin(langs: _*))
+      .extractUASTs
+      .select("repository_id", "path", "blob_id", "uast")
+      .filter(_.getAs[Seq[Array[Byte]]]("uast").nonEmpty)
+      .withColumn("document", expr("CONCAT(repository_id, '//', path, '@', blob_id)"))
+      .select("document", "uast")
+  }
+
+  def sparkFeatures(uasts: DataFrame): RDD[SparkFeature] = {
```
bzz: Just wondering, if it would be worth adding a few lines of comment here, explaining why DataFrame API \w table-like structure that is used in the rest of the code is not a good fit and RDD of case class is used instead.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -413,7 +417,6 @@ object Gemini {
   val defaultBblfshPort: Int = 9432
   val defaultFeHost: String = "127.0.0.1"
   val defaultFePort: Int = 9001
-  val defaultDocFreqFile: String = "docfreq.json"
```
bzz: I see big benefit in having all the configuration defaults in the single place, do you think it might be worth keeping it that way for now? Then we could get back to this single place and clean it up.

src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
```diff
@@ -126,6 +127,9 @@ class CassandraSparkSpec extends FlatSpec
 
   object Cassandra extends Tag("Cassandra")
 
+  // such tests require bblfsh & feature extractors
+  object Integration extends Tag("Integration")
```
bzz: 👍, why not calling  it
```scala
object RequireBblfshAndFeatureExtractor
```
? 

That way reader of https://github.com/src-d/gemini/pull/105/files#diff-354f30a63fb0907d4ad57269548329e3R22 has additional benefit of not needing to think what excatly does "Integration" mean in this context.

src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
```diff
@@ -241,10 +245,65 @@ class CassandraSparkSpec extends FlatSpec
 
   "Hash with limit" should "collect files only from limit repos" in {
     val gemini = Gemini(sparkSession)
-    val repos = gemini.hash("src/test/resources/siva", 1).select(Gemini.meta.repo).distinct().count()
+    val repos = gemini.hash("src/test/resources/siva", 1).select("repository_id").distinct().count()
```
bzz: Why is it needed to stop using the variable in favor of the constant?

Ideally, we should not have any string literals in for the fields and have `table["meta"].repo`, `table["hashtable"].something`, etc. This simplifies refactorings and increase maintainability.

It most probably will be done as a separate PR with refactoring later on, but until then, if there is no strong reason, I suggest to keep using this var at least for a 'meta' table.

smacker: because now hash doesn't return DF for database anymore. But original DF of engine spark. We have to keep original names to be able to call UDFs.

src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
smacker: Integration it a common term which means tests require external dependencies.
I don't think `RequireBblfshAndFeatureExtractor` is a good name for it.
We may remove FeatureExtractor service (implementing it in scala) or add one more external dependency (community-detection-as-a-service for example). It shouldn't cause renaming of this tag.
Also it is `RequireCassandaAndBblfshAndFeatureExtractor` which doesn't look nice.
I really don't want it to become `RequireCassandaAndBblfshAndFeatureExtractorAndAnotherServiceAndOneMoreAndAnotherOne` later.

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: Mostly because I had problems with implementing it with DataFrame and Apollo use RDD here.
But I agree it's better to use DataFrame. Let me come back to it in sequential PR. For now, I added TODO.

src/main/scala/tech/sourced/gemini/Gemini.scala
smacker: makes sense. Fixed.

project/Dependencies.scala
smacker: sorry. leftover. that's why PR is WIP :)
rename of rename is done.

smacker:
All feedback is addressed (mostly fixes and 2 comments).
I'm merging it as soon as CI is green.

The next PR changes code quite much, so if you still see any issues here, they will be addressed there.

src/test/scala/tech/sourced/gemini/CassandraSparkSpec.scala
bzz: > Integration it a common term which means tests require external dependencies.

True, but it's too broad an generic and depends on context a lot. I find readable variable names very useful, much more then comments near them and it's totally fine if sometimes things need to be renamed though refactoring.

From my side - a productive discussion here would include debating alternatives, rather then the approach though.

Issue #107, opened 2018-04-24
Speedup CI


smacker:
The longest task is integration. Most of the time takes:

- install python. We can improve it a bit, installing some deps from apt instead of building
- docker for FE. We can remove it now when we have python in CI
- Make build. We can try to use travis cache to improve it

also check that we don't run services when they aren't necessary

PR #109, opened 2018-04-24
Speedup CI

smacker:
Fixes: #107 

the time, of course, depends on the network. But.

Master:
master

No cache:
no-cache

Cached:
cached

**Details:**

ubuntu trusty doesn't have packages for python3 that we need :(

python deps: ~25s
scala build time: ~150s now. Locally it's even slower
python deps for fe instead of docker: ~35s
db starts very slow ~1m.
no more travis cache update if there is nothing to change

pip http requests have an expiration time, so it's faster to download them again than rebuild&upload travis cache every time.

**Conclusion:**
It might be possible to improve time for python dependencies (reduce 1min), but everything else works almost as fast as possible.

### DIFF
--- a/scripts/start_docker_bblfsh.sh
+++ b/scripts/start_docker_bblfsh.sh
@@ -8,19 +8,20 @@ image="bblfsh/bblfshd:v2.4.2"
 
 set -ev
 
-if [[ -z "${STYLE_CHECK}" ]]; then
-  echo "Starting Docker image ${image}"
-  mkdir -p $HOME/bblfsh-drivers
-  docker run -v $HOME/bblfsh-drivers:/var/lib/bblfshd --name bblfshd --privileged -p 9432:9432 -d "${image}"
+echo "Starting Docker image ${image}"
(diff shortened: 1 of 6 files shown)


scripts/start_docker_bblfsh.sh
```diff
@@ -8,19 +8,15 @@ image="bblfsh/bblfshd:v2.4.2"
...
-  docker exec -it bblfshd bblfshctl driver install --recommended
-else
-  echo "Skip starting Docker image ${image}"
+if [[ "$?" -ne 0 ]]; then
+  echo "Unable to start Docker ${file_list}" >&2
+  exit "${E_BAD_DOCKER}"
 fi
 
+docker exec -it bblfshd bblfshctl driver install --recommended
+
```
bzz: While we are still at it, to make it bullet-poof, why do not we add here [return value check](https://google.github.io/styleguide/shell.xml?showone=Checking_Return_Values#Checking_Return_Values), same as for `docker run` above?

smacker:
check added

Issue #55, opened 2018-03-14
Hash: add similar files


bzz:
Umbrella issue for batch-hashing similar files on Apache Spark:

 - UAST, feature extraction (same args as in query)
 - Tokenize, Vectorize: wBOW, tf/idf - defining a pipeline 
 - Use CPU-based WMH implementation in Scala (same args as in query)
 - Take results of hash pipeline, create hash tables, write to ScyllaDB \w schema as in Apollo
 - Test correctness: Apollo and Gemini produces same results (use same seed if needed)

bzz:
Updated description above changed
```
    - write docFreq to database (if we write in the driver, if we are able to write in workers it might be simpler to use parquet file on distributed fs)
    - change Query to consume docFreq from database/parquet and calculate params based on a seed
```
to writing to JSON files first that query is already knows how to consume, and added a new subsequent task for switching to DB on both sides.

Rationale is: on the query side, params should be used exactly the same as on hash side so we should save them. JSON serialization is generic enough and is already working on query side, so first implementation of hash should also use it. Later this could be change do DB with well-defined schema.

bzz:
For Gemini, as we use CPU hashing, we do not need to batch the bags before hashing.

smacker:
@bzz that image is a bit outdated. Here is a new one for apollo:

![hash](https://user-images.githubusercontent.com/406916/39181703-c7cbf120-47ba-11e8-9f14-2cfe0155974d.png)

bzz:
Updated the description and moved out saving DocFreq to DB to a separate issue #112 to decouple this work.

PR #111, opened 2018-04-27
Implement hash similarity + refactoring

smacker:
Part of https://github.com/src-d/gemini/issues/55

First of all sorry for a huge bunch of code. I had reasons, it was just too difficult to develop as before. You can see them in the description of changes below:

- Implement `hash` in gemini
- Remove support for external params.json & hashtables-num/band-size (to make sure hash&query use the same options)
- Extract `query` and `hash` related methods to separate classes, common code (wmh related functions) to another one. It became too difficult to follow code logic otherwise because those commands do similar things but non-spark/spark way.
- Make most of the methods for query&hash protected due to their dependencies on each other
- Running all test suite as before became way too slow for developing:
  - Tests are split to separate suits by command (we can run them now using `./sbt test-only <suite-name>`)
  - Added tags for each dependency (solves the problem from the previous PR with `VeryLongNameForTag`)
  - Removed dependency on hash from query tests (we have integration tests using commands for it)
  - We don't need to test commands 2 times in cluster mode and a regular one. Just a cluster one would be enough.

The PR is way too huge already, but later I would prefer to make the same refactoring for report command to speedup CI.


bzz:
I belive this is also a part of the #55 ☂️  scope

.travis.yml
```diff
@@ -28,17 +29,12 @@ matrix:
...
+      - tar -xzf src/test/resources/weighted-minhash/csv.tar.gz -C src/test/resources/weighted-minhash/
     script:
       # start dependencies
       - VIRTUAL_ENV=None IN_BACKGROUND=1 ./feature_extractor
-      # just check that commands don't fail
-      - ./hash src/test/resources/siva || travis_terminate 1
-      - ./query ./src/test/resources/LICENSE
-      - ./report
-      - go get ./src/main/go/... || true
-      - go run ./src/main/go/query.go ./src/test/resources/LICENSE
```
bzz: Sorry if I missed something, but what is the reason to remove tests for go-based duplication detection?

It might be ok to move them to different profile to optimize CI speed, or move to the next one, with external Apache Spark, but I would prefer keeping those somewhere.

.travis.yml
```diff
@@ -74,18 +70,14 @@ matrix:
...
-      - ./sbt "test-only * -- -n Integration"
       # hashing test repositories
       - ./hash src/test/resources/siva || travis_terminate 1
       # query without similarity
       - ./query ./src/test/resources/LICENSE
       # report without similarity
       - ./report
       # query with similarity
-      - docker cp ./src/test/resources/hashtables_dump.cql db:/hashtables_dump.cql
-      - docker exec -it db cqlsh -e "SOURCE '/hashtables_dump.cql'"
```
bzz: No need for hashtables fixtures any more?

src/main/scala/tech/sourced/gemini/Database.scala
```diff
@@ -0,0 +1,46 @@
...
+  /**
+    * query database by hash and return files with that hash
+    *
+    * @param sha
+    * @param conn
+    * @param keyspace
+    * @param tables
+    * @return
+    */
+  def repoFilesByHash(sha: String, conn: Session, keyspace: String, tables: Tables): Iterable[RepoFile] = {
```
bzz: Just wondering if `findFilesByHash()` be more readable in this case.

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -0,0 +1,153 @@
...
+
+    log.info(s"${similarShas.length} SHA1's found to be similar")
+
+    val similar = similarShas.flatMap(sha1 => Database.repoFilesByHash(sha1, conn, keyspace, tables))
+    QueryResult(duplicates, similar)
+  }
+
+  protected def findDuplicatesOfFile(file: File): Iterable[RepoFile] = {
+    Database.repoFilesByHash(Gemini.computeSha1(file), conn, keyspace, tables)
+  }
```
bzz: It seems to me that this guy can be safely removed in favor of `Database.findFilesByHash()` or something like that.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+                    features: Iterable[Feature],
+                    sampleSize: Int = defaultSampleSize,
+                    seed: Int = defaultSeed): Array[Array[Long]] = {
+    val OrderedDocFreq(docs, tokens, df) = docFreq
+
+    if (wmh == null || tokensSize != tokens.size) {
+      wmh = new WeightedMinHash(tokens.size, sampleSize, seed)
+    }
+
+    val bag = mutable.ArrayBuffer.fill(tokens.size)(0.toDouble)
```
bzz: For me, ability to understand this logic would increase, if extracted to a method with descriptive name and possibly a link to analog in Apollo codebase in ScalaDoc comment.

src/main/scala/tech/sourced/gemini/Database.scala
```diff
@@ -0,0 +1,46 @@
...
+    val cols = tables.metaCols
+    val query = QueryBuilder.select().all().from(keyspace, tables.meta)
+      .where(QueryBuilder.eq(cols.sha, sha))
+
+    conn.execute(query).asScala.map { row =>
+      RepoFile(row.getString(cols.repo), row.getString(cols.commit), row.getString(cols.path), row.getString(cols.sha))
```
smola: You could move the row-to-repofile conversion to a separate method.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+        case _ =>
+      }
+    }
+    wmh.hash(bag.toArray)
+  }
+
+  // split hash to bands
+  // a little bit verbose because java doesn't have uint32 type
+  // in apollo it's one-liner:
+  // https://github.com/src-d/apollo/blob/57e52394783d73e38cf1f862afc0166724991fd5/apollo/query.py#L35
```
bzz: 👍 It could be even nicer, if converted to a single ScalaDoc comment.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
+package tech.sourced.gemini
+
+import tech.sourced.featurext.generated.service.Feature
+
+import scala.collection.mutable
+
+object FeaturesHash {
+  // All params below must be the same for hash & query
+  val defaultSeed = 0
```
smola: Constants usually use `CamelCase` if using [Scala-style](https://docs.scala-lang.org/style/naming-conventions.html) or `UPPER_CASE` if using Java-style (which some people uses in Scala too).

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+  val defaultSampleSize = 128
+  // Depend on threshold and sample_size
+  // this values calculated for threshold=0.8 and sample_size=128
+  val defaultHashtablesNum = 9
+  val defaultBandSize = 13
+
+  private var tokensSize: Int = _
```
smola: When is `tokensSize` initialized?

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+  private var wmh: WeightedMinHash = _
+
+  def hashFeatures(
+                    docFreq: OrderedDocFreq,
+                    features: Iterable[Feature],
+                    sampleSize: Int = defaultSampleSize,
+                    seed: Int = defaultSeed): Array[Array[Long]] = {
+    val OrderedDocFreq(docs, tokens, df) = docFreq
+
+    if (wmh == null || tokensSize != tokens.size) {
```
smola: `tokensSize != tokens.size` seems to be always true?

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -0,0 +1,153 @@
...
+          val featuresList = extractFeatures(feClient, node)
+          findSimilarFiles(featuresList, conn, docFreqFile)
+        case _ => Seq()
+      }
+    }
+  }
+
+  private def findSimilarFiles(featuresList: Iterable[Feature], conn: Session, docFreqFile: File): Seq[String] = {
+    if (featuresList.isEmpty) {
+      log.warn("file doesn't contain features")
```
bzz: Just for the sake of ability to debug it from the logs, let's include a particular filename in this message.
Could be done either by passing a original filename in this function or by pushing the check to the client.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smola: What's expected to be bigger, `features` or `tokens`?

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+          bag(idx) = MathUtil.logTFlogIDF(tf, df(feature.name), docs)
+        }
+        case _ =>
+      }
+    }
+    wmh.hash(bag.toArray)
+  }
+
+  // split hash to bands
+  // a little bit verbose because java doesn't have uint32 type
```
smola: Comments about implementation are better inside the function body itself, while preserving the comment on top of the method declaration for scaladocs.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,62 @@
...
+  // https://github.com/src-d/apollo/blob/57e52394783d73e38cf1f862afc0166724991fd5/apollo/query.py#L35
+  def wmhToBands(
+                  wmh: Array[Array[Long]],
+                  htnum: Int = defaultHashtablesNum,
+                  bandSize: Int = defaultBandSize): Iterable[Array[Byte]] = {
+    (0 until htnum).map { i =>
+      val from = i * bandSize
+      val to = (i + 1) * bandSize
+      val band = (from until to).foldLeft(Array.empty[Byte]) { (arr, j) =>
+        arr ++ MathUtil.toUInt32ByteArray(wmh(j)(0)) ++ MathUtil.toUInt32ByteArray(wmh(j)(1))
```
smola: If using an array where you know the size upfront, it is more efficient to create the array of the desired size and fill it later.

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -0,0 +1,153 @@
...
+
+    val similar = similarShas.flatMap(sha1 => Database.repoFilesByHash(sha1, conn, keyspace, tables))
+    QueryResult(duplicates, similar)
+  }
+
+  protected def findDuplicatesOfFile(file: File): Iterable[RepoFile] = {
+    Database.repoFilesByHash(Gemini.computeSha1(file), conn, keyspace, tables)
+  }
+
+  protected def findSimilarForFile(
```
smola: Maybe you want to put these methods in a companion object, if they do not use any class field.

src/main/scala/tech/sourced/gemini/Hash.scala
```diff
@@ -0,0 +1,171 @@
...
+      .getHEAD
+      .getCommits
+      .getTreeEntries
+      .getBlobs
+      .filter('is_binary === false)
+
+  protected def extractUast(files: DataFrame): DataFrame = {
+    // TODO(max): get languages from bblfsh directly as soon as
+    // https://github.com/bblfsh/client-scala/issues/68 resolved
+    val langs = Array("Java", "Python", "Go", "JavaScript", "TypeScript", "Ruby", "Bash", "Php")
```
smola: Prefer `Seq` by default, unless you actually want a specific implementation and use `Array`, `List`, etc.

src/test/scala/tech/sourced/gemini/BaseDBSpec.scala
```diff
@@ -0,0 +1,55 @@
+package tech.sourced.gemini
+
+import com.datastax.driver.core.{Cluster, Session}
+import org.scalatest.{BeforeAndAfterAll, Suite, Tag}
+
+case class hashtableItem(hashtable: Int, v: String, sha1: String)
```
smola: The Go convention of lower-case for private stuff does not apply to Scala ;-) https://docs.scala-lang.org/style/naming-conventions.html

.travis.yml
smacker: we still run go-cli and all others here:
https://github.com/smacker/gemini/blob/3ac7c7630a3c5a8912e5d0735cc3b30a7b8ee66f/.travis.yml#L89

I don't see any benefits to run it on local spark first and then on spark in cluster mode. Imo, cluster mode should be enough.

.travis.yml
smacker: No. Hash above produces hashtables now.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smacker: could you help me to write comment/change name?

The method has name `hashFeatures` it's exactly what this code is doing.
I also tried to write it not short but understandable giving variable meaningful names like bag/tf and so on. So I don't really see how to improve it by myself.

The code in apollo: https://github.com/src-d/apollo/blob/master/apollo/hasher.py#L204
but I personally get confused by it.

src/main/scala/tech/sourced/gemini/FileQuery.scala
smacker: Ok. But sorry what logs? It's query. You run it as cli and pass file there.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smacker: @bzz should we change it everywhere?

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -0,0 +1,153 @@
...
+      Seq()
+    } else {
+      val cols = tables.hashtablesCols
+      val wmh = hashFile(featuresList, docFreqFile)
+
+      val bands = FeaturesHash.wmhToBands(wmh)
+
+      log.info("Looking for similar items")
+      val similar = bands.zipWithIndex.foldLeft(Set[String]()) { case (sim, (band, i)) =>
+        val table = "hashtables"
```
carlosms: Should this be tables.hashtables?

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smacker: wow. Thanks a lot!

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smacker: it shouldn't be after fix of commend above.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smacker: tokens (vocabulary of all features) in most cases will be much-much-much bigger than features (for a single file)

src/main/scala/tech/sourced/gemini/FileQuery.scala
smacker: it uses log from a class instance right now. But I see I can remove some params from this method cause they belong to the class instance too. Thanks!

src/main/scala/tech/sourced/gemini/FileQuery.scala
smacker: thanks! missed!

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smola: If you end up deciding to change it everywhere, it's probably better to do it on its own PR.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
smola: Ok, so if `tokens` is an `Array[String]` and it is guaranteed to be sorted, you can improve performance by replacing the `indexOf` call with something like `java.util.Arrays.binarySearch(a.asInstanceOf[Array[AnyRef]],key)`.

.travis.yml
bzz: On Go 👍 , on Cluster mode - avoid having Spark Standalone installed/unpacked would just cut some CI some.

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
bzz: If nobody has a strong opinion about this, I belive it is good as-is and there is no need to change anything as it follows the styleguide:

> That is, each word is capitalized, except possibly the first word

smacker:
@bzz another pass, please.
@smola if you have time, please check if I applied your comments properly

src/main/scala/tech/sourced/gemini/Hash.scala
```diff
@@ -0,0 +1,172 @@
...
+    val cols = tables.hashtablesCols
+    rdd.flatMap(row => {
+      val RDDHash(doc, wmh) = row
+      FeaturesHash.wmhToBands(wmh).zipWithIndex.map{ case(band, i) => (doc, i, band) }
+    })
+      .toDF(cols.sha, cols.hashtable, cols.value)
+      .write
+      .mode("append")
+      .cassandraFormat(tables.hashtables, keyspace)
+      .save()
```
bzz: A nitpick, but this formatting for method-chaining looks more consistent

```scala
    rdd
      .flatMap { row =>
        val RDDHash(doc, wmh) = row
        FeaturesHash.wmhToBands(wmh).zipWithIndex.map { case(band, i) => (doc, i, band) }
      }
      .toDF(cols.sha, cols.hashtable, cols.value)
      .write
      .mode("append")
      .cassandraFormat(tables.hashtables, keyspace)
      .save()
```

src/test/scala/tech/sourced/featurext/ClientSpec.scala
```diff
@@ -24,47 +25,55 @@ class ClientSpec extends FlatSpec
...
   }
 
+  override def afterAll(): Unit = {
+    channel.shutdownNow()
+  }
+
   "identifiers call" should "return correct response" in {
     val request = IdentifiersRequest(docfreqThreshold=5, uast=Some(uast))
     val reply = blockingStub.identifiers(request)
+    var features = reply.features.sortBy(f => f.name)
```
bzz: same ✨ 
```
.sortBy(_.name)
```
if 😎

src/test/scala/tech/sourced/gemini/BaseSparkSpec.scala
```diff
@@ -2,32 +2,50 @@ package tech.sourced.gemini
...
+      .set("spark.cleaner.ttl", "3600")
+
+    useSparkConf(defaultConf)
+  }
+
   override protected def beforeAll(): Unit = {
     super.beforeAll()
 
     sparkSession = SparkSession.builder()
-      .appName("test").master("local[*]")
```
bzz: I find it useful, in case Apache Spark is configued to preserve history, to be able to identify sessions that belong to tests, and simplest way to do it in WebUI is by name. 

May be there can be a better name, but do you see any particular reason to remove this?

src/test/scala/tech/sourced/gemini/BaseSparkSpec.scala
```diff
@@ -2,32 +2,50 @@ package tech.sourced.gemini
...
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
-import org.scalatest.{BeforeAndAfterAll, Suite}
+import org.scalatest.{BeforeAndAfterAll, Suite, Tag}
 
 trait BaseSparkSpec extends BeforeAndAfterAll {
   this: Suite =>
 
-  @transient var sparkSession: SparkSession = _
+  var sparkSession: SparkSession = _
```
bzz: Could you please remind, what is the reason to change it here?

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -0,0 +1,75 @@
...
+  private var wmh: WeightedMinHash = _
+
+  def hashFeatures(
+                    docFreq: OrderedDocFreq,
+                    features: Iterable[Feature],
+                    sampleSize: Int = defaultSampleSize,
+                    seed: Int = defaultSeed): Array[Array[Long]] = synchronized {
+    val OrderedDocFreq(docs, tokens, df) = docFreq
+
+    if (wmh == null || tokensSize != tokens.size) {
```
smola: This means that `FeaturesHash` is not thread-safe. If that is intended, it would be good to add a `scaladoc` to the `object` and warn about it.

smacker: It's `synchronized` method... Do I still need to document it in scaladoc?

src/test/scala/tech/sourced/gemini/BaseSparkSpec.scala
smacker: there is `setAppName(this.getClass.getSimpleName)` in `useDefaultSparkConf` now.
It will produce different names for different test suites.

And in general set the name in config looks cleaner.

src/test/scala/tech/sourced/gemini/BaseSparkSpec.scala
smacker: the only reason is to simplify the code.
I couldn't find why we need to set spark session unserializable here, so looks like we don't need.
To avoid wasting the time of the next developer who stumble onto it I removed it.

src/main/scala/tech/sourced/gemini/Hash.scala
smacker: fixed

src/test/scala/tech/sourced/gemini/BaseSparkSpec.scala
bzz: I see. I would rather keep it transient, as it follow the same pattern as tests in [Apache Spark do](https://github.com/apache/spark/blob/ceaf77ae43a14e993ac6d1ff34b50256eacd6abb/mllib/src/test/scala/org/apache/spark/mllib/util/MLlibTestSparkContext.scala#L30) - it seems have to do something with ability to make sure that there is only one SparkContext in JVM.

Not sure though, how that could happen in our case, but at least for the sake of [consistency](https://github.com/holdenk/spark-testing-base/blob/c6b4c74ea6274b2a13c9226dbdb82f435387b60d/src/main/2.0/scala/com/holdenkarau/spark/testing/DataFrameSuiteBase.scala#L57) I would keep it as is.

smacker: reverted. But I don't like having magic copy-past without understanding why we need it.

Issue #95, opened 2018-04-12
Add Docker compose for dependencies


bzz:
To streamline first user experience:
 - add simple Docker compose script for dependencies: DB and FE - it will enable part of #84 
 - make CI use it

PR #115, opened 2018-05-07
add docker-compose

smacker:
Part of #95 

With this docker-compose.yml it's possible to do:

- `docker-compose up -d scylla feature_extractor`: would run deps for gemini on local machine
- `docker-compose up -d` && `docker-compose exec gemini ./report` run report command without any dependencies on host machine.

CI will be updated in another commit (after merge of #111)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,46 @@
+version: "3"
+services:
+  gemini:
+    build: .
+    volumes:
+      - ./repositories:/repositories
+    network_mode: "host"
+
+  scylla:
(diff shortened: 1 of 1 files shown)


bzz:
Looks great, why not to also include in this PR an update to https://github.com/src-d/gemini/#run that adds subsections:

```
## Run
### Standalone
<old content of Run section here>

### With Docker
<instructions on docker-compose from this PR>
```

What do you think?

smacker:
Documentation will be updated in another PR with adding env variables (at least for now in bash scripts) to avoid problems with busy ports.

Issue #113, opened 2018-05-04
File-similarity: apply default parameters, found by ML team


bzz:
ML team [has done extensive research](https://github.com/src-d/backlog/issues/1197#issuecomment-383632529) on number of optimal parameters for file-based similarity detection:
 - best feature extractors for this job
 - feature extractors weights
 - threshold
 - hash size

@EgorBu could you please take a look and help by:
 - check that no parameters are missing from the list above
 - provide a default values for those, that we should be using in Gemini

EgorBu:
parameters for val_acc 0.9497 - best so far
* best combination of feature extractors`('lit', 'id', 'graphlet')`
* `id_split_stem True` - identifiers should be splitted
* `lit_weight 264.38490183192016`
* `id_weight 194.2959667378745`
* `graphlet_weight 547.7754676767302`
* `threshold 0.6456707644812872`
* `hash size 160`

PR #116, opened 2018-05-07
Apply default params

smacker:
Fix: #113 
Based on: #111 

Update is here: https://github.com/src-d/gemini/commit/e83e750a83d392e569acb625ac354dd2312d332d
Small script to calculate hashtables params: https://github.com/src-d/gemini/commit/311afe001f1c38848751d9a8dff7b0437d59b23b
Fixtures are also updated using new defaults.


bzz:
Could you check if `./sbt scalastyle` and `make lint-python` pass as well?

Also, could you please document a bit, how `calculate_hashtables/main.py` is intended to be used?  https://github.com/src-d/gemini#dev could be a good place

src/main/python/calculate_hashtables/main.py
```diff
@@ -0,0 +1,61 @@
+"""
+Calculate number of hashtabels and size of a band
+using the threshold and sample size of wmh
+"""
```
bzz: If this code is similar or comes from another codebase, could you please add a permlink URL in here as well?

I.e https://github.com/src-d/apollo/blob/bbb9f83fffc93e791fa59f064efadefa270f400a/apollo/hasher.py#L20

bzz: As this code seems to come from [MIT-licensed project](https://github.com/ekzhu/datasketch/blob/master/LICENSE), so you also need to update https://github.com/src-d/gemini/blob/master/LICENSE#L677 by adding a new [section for MIT](https://github.com/apache/spark/blob/master/LICENSE#L276) and then as before - add a link to both, original project and a path to `main.py` in our codebase that contains this code.

smacker:
Not sure we should document this script in readme. It calculates hard-coded params, so it meant to be used only internally. Docstring in the file itself updated.

bzz:
Got it. Internally though, could you help understanding, in what case one would want to run it?

smacker:
We may want to rerun it if ML team provide us new hyper-params.

PR #117, opened 2018-05-08
(almost) single command for gemini

smacker:
In this PR:

- update `HashSparkApp` to accept keyspace/bblfsh address/features extractor address same as `query` does
- update bash scripts to read connect parameters from env vars.
- update default `docker-compose` to use internal network
- move host port bindings to separate `docker-compose`
- update README with instruction how to run gemini using docker-compose. (other sections **AREN'T UPDATED**, it will be done in separate PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,96 @@ Supported granularity level or items are:
  - files
  - functions (TBD)
 
+## Run
+
+```
+./hash   <path-to-repos-or-siva-files>
+./query  <path-to-file>
+./report
+```
+
+You would need to prefix commands with `docker-compose exec gemini` if you run it in docker. Read below how to start gemini in docker or standalone mode.
+
+### Hash
+To pre-process number of repositories for a quick finding of the duplicates run
+
+```
+./hash ./src/test/resources/siva
+```
+
+Input format of the repositories is the same as in [src-d/Engine](https://github.com/src-d/engine).
+
+### Query
+To find all duplicate of the single file run
+
+```
+./query <path-to-single-file>
+```
+
+There is an example of a client in a Golang under [src/main/go](src/main/go)
+
+### Report
+To find all duplicate files in all repositories run
+
+```
+./report
+```
+
+All repositories must be [hashed](#hash) before.
+
 ## Requirements
+
+### Docker
+
+Start containers:
+
+```bash
+docker-compose up -d
+```
+
+Local directories `repositories` and `query` are available as `/repositories` and `/query` inside the container.
+
+Examples:
+
+```bash
+$ docker-compose exec gemini ./hash /repositories
+Hashing 2 repositories in: /repositories
+	file:/repositories/f281ab6f2e0e38dcc3af05360667d8f530c00103.siva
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -82,6 +44,80 @@ To find all duplicate files in all repositories run
...
+docker-compose up -d
+```
+
+Local directories `repositories` and `query` are available as `/repositories` and `/query` inside the container.
+
+Examples:
+
+```bash
+$ docker-compose exec gemini ./hash /repositories
+Hashing 2 repositories in: /repositories
```
bzz: 👍  for detailed example that users could just copy and run.

I would suggest  to remove the output of those commands here and below for 2 reasons:
 - easier to maintain, i.e when format of output is changed - no need to update this
 - consistency, as other examples above do not include it

README.md
```diff
@@ -82,6 +44,80 @@ To find all duplicate files in all repositories run
...
+	file:/repositories/f281ab6f2e0e38dcc3af05360667d8f530c00103.siva
+	file:/repositories/9279be3cf07fb3cca4fc964b27acea57e0af461b.siva
+Done
+
+$ docker-compose exec gemini ./query /query/consumer.go
```
bzz: Question: would that be possible to use single `docker-compose` just for running all the dependencies together and then keep examples here with just a `docker run`?

Asking, as I think from the product standpoint that may be conceptually a bit simple for some of the users.

Or do you think it would have some technical downsides?

scripts/common_params.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+# Parse parameters for dependencies or use environment variable
+#
+# List of params is common for all commands:
+# Database: DB_HOST/DB_PORT
+# Bblfshd: BBLFSH_HOST/BBLFSH_PORT
+# Feature extractor: FE_HOST/FE_PORT
```
bzz: 👍 for extracting this logic to re-usable component.

Quick question - as hash, query and report all use it, could you please help me understand what would be the possible downsides of having it implemented directly on JVM application side?

README.md
smacker: do you mean replace `docker-compose exec` with `docker exec`?
It wouldn't be simpler because the name of the container is something like `gemini_gemini_1`. And the last number can be different.

docker-compose.host.yml
```diff
@@ -0,0 +1,16 @@
+version: "3"
```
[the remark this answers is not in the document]
smacker: default docker-compose file is meant to be used by a user.

`docker-compose.host.yml` uses host network and requires free ports, it is meant to be used by CI or developer.
Having separate config files is the only way to configure docker-compose for different cases.

Here is a documentation for it: https://docs.docker.com/compose/extends/

README.md
smacker: removed

bzz:
Looks great to me, thank you @smaker!

smacker:
Issue #120 create. Merging.

scripts/common_params.sh
bzz: Assume that @smacker has just forgot to do it https://github.com/src-d/gemini/issues/120

PR #118, opened 2018-05-08
update CLI arguments documentation

smacker:
This PR documents CLI arguments.

Some internal arguments such as docfreq file path or connected components output dir left undocumented on purpose.

There is also `mode` argument for report, but I don't really understand why we need it and it's not going to work by default. (default DB is scylla, not Cassandra for now) So I left it undocumented too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,8 +108,8 @@ For `query` and `hash` commands parameters for bblfsh/features extractor  config
 
 Hash command specific arguments:
 
- * `-l/--limit` - limit on the number of processed repositories, no default
- * `-f/--format [siva | bare | standard]` - format of the stored repositories, default `siva`
+ * `-l/--limit` - limit the number of repositories to be processed. All repositories will be processed by default
+ * `-f/--format` - format of the stored repositories. Supported input data formats that repositories could be stored in are `siva`, `bare` or `standard`, default `siva`
 
 # Dev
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,9 +92,24 @@ MASTER="spark://<spark-master-url>" ./hash <path>
...
+For `query` and `hash` commands parameters for bblfsh/features extractor  configuration are available:
+
+ * `--bblfsh-host` - babelfish server host, default `127.0.0.1`
+ * `--bblfsh-port` - babelfish server port, default `9432`
+ * `--features-extractor-host` - features-extractor host, default `127.0.0.1`
+ * `--features-extractor-port` - features-extractor port, default `9001`
+
+Hash command specific arguments:
+
+ * `-l/--limit` - limit on the number of processed repositories, no default
```
bzz: -> `limit the number of repositories to be processed. All repositories will be processed by default`

README.md
```diff
@@ -92,9 +92,24 @@ MASTER="spark://<spark-master-url>" ./hash <path>
...
+ * `-f/--format [siva | bare | standard]` - format of the stored repositories, default `siva`
```
bzz: Let's be explicit that it's about _input_ data format here by updating a description to something like 
`supported input data formats that repositories could be stored in, ...`

bzz:
Looks awesome!

PR #119, opened 2018-05-09
use docker-compose in CI

smacker:
based on #117

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
   include:
   - scala: 2.11.2
     before_install:
-    - ./scripts/start_docker_db.sh
+    - docker-compose -f docker-compose.yml -f docker-compose.host.yml up -d scylla
     - tar -xzf src/test/resources/weighted-minhash/csv.tar.gz -C src/test/resources/weighted-minhash/
     script:
       # run only tests without external deps (beside DB)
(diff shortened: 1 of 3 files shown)


scripts/start_docker_bblfsh.sh
```diff
@@ -1,27 +0,0 @@
-#!/bin/bash
```
bzz: I find those 3 to be useful examples of how to run containers with all the options, etc individually with just `docker` that i.e may be useful for local debugging.

Do you think we could keep them around and delete, when this knowledge get's transferred to the  a Developer's docs? 

It's nice to have `docker-compose` to run all of them and I know that all the options, etc for individual docker command can be recovered from compose configuration, but I still see value for having those individual commands in some place.

smacker: don't see any value to keep scripts as documentation. but whatever.

bzz:
Assume there still might be some WIP as CI seems to be red.

smacker:
@bzz CI was red due to some internal travis problems. (tasks just hanged and died by timeout)
I have restarted it and everything is green again.

smacker:
github didn't recognize that commits are in master already.

I have rebased just to be sure everything is ok.
Going to merge as CI pass again.

Issue #120, opened 2018-05-10
Refactoring: handle all CLI in common opts parser


smacker:
1. Apps parsers should inherit base parser with common options like db, versbose and so on.
2. Add support to read the same options from env variables

bzz:
This will allow to get rid of https://github.com/src-d/gemini/blob/master/scripts/common_params.sh

bzz:
Do not forget: change shell wrappers to remove this https://github.com/src-d/gemini/pull/125/files#r194036469

bzz:
and also remove it from `./release.sh`

PR #125, opened 2018-06-08
Packaging

### LINKED ISSUE #124: Improve packaging to reduce applications start time
Right now all commands rely on running `sbt` to get the right set of dependencies at runtime classpath.

Also only Spark application's release artefact for `./hash` is a fatJar and include all dependencies (except Spark itself), and `./query` and `./report` are slim and rely on presence of dependencies at classpath from external source.

This issue is about changing the build structure, so:
 - Spark and non-Spark applications build artefacts ideally need to be fat jars with all the dependencies included
 - shell wrappers need to be changed to become something like `java -cp XXX ClassName`

This will still have overhead of JVM process cold start time, but that is totally fine for now.

bzz:
Fixes #124 

- [x] better CLI output on -v
- [x] fatJar packaging
- [ ] fatJar + deps

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,5 +157,5 @@ bblfsh-sdk-tools fixtures -p .proto -l <LANG> <path-to-source-code-file>
 
 ## License
 
-Copyright (C) 2017 source{d}.
+Copyright (C) 2018 source{d}.
 This project is licensed under the [GNU General Public License v3.0](LICENSE).
--- a/README.md
+++ b/README.md
@@ -135,11 +135,6 @@ Build, fatJar for Apache Spark (hash, report)
 ./sbt assembly
 ```
 
-Build, for query
-```
-./sbt package
-```
-
 Tests (with embedded Cassandra)
 ```
 ./sbt test
--- a/hash
+++ b/hash
@@ -11,7 +11,7 @@ deps_jar="target/gemini-assembly-${version}-deps.jar"
 build_deps_command="./sbt assemblyPackageDependency"
 current_dir="$(dirname "$0")"
 
-app_class="tech.sourced.gemini.HashSparkApp"
+app_class="tech.sourced.gemini.cmd.HashSparkApp"
 app_name="Gemini - hashing"
 
 hash java >/dev/null 2>&1 || { echo "Please install Java" >&2; exit 1; }
--- a/hash
+++ b/hash
@@ -29,6 +29,7 @@ fi
 
 sparkSubmit() {
     if hash spark-submit 2>/dev/null; then
+        echo "Using spark-submit from PATH, which is $(which spark-submit)"
         exec spark-submit "$@"
     elif [[ -n "${SPARK_HOME}" ]]; then
         echo "Using spark-submit from ${SPARK_HOME}"
--- a/hash
+++ b/hash
@@ -1,14 +1,14 @@
 #!/bin/bash
-# Convenience script to run hashing over a set repositories in .siva files
+# Convenience script to run hashing over a set repositories
 
 E_NO_SPARK=2
 E_BUILD_FAILED=1
 
-version=$(grep "version :=" build.sbt | cut -d'=' -f 2 | tr -d ' ' | tr -d '"')
+#version=$(grep "version :=" build.sbt | cut -d'=' -f 2 | tr -d ' ' | tr -d '"')
 jar="target/gemini-uber.jar"
 build_command="./sbt assembly"
-deps_jar="target/gemini-assembly-${version}-deps.jar"
-build_deps_command="./sbt assemblyPackageDependency"
+#deps_jar="target/gemini-assembly-${version}-deps.jar"
+#build_deps_command="./sbt assemblyPackageDependency"
 current_dir="$(dirname "$0")"
 
 app_class="tech.sourced.gemini.cmd.HashSparkApp"
@@ -16,15 +16,18 @@ app_name="Gemini - hashing"
 
 hash java >/dev/null 2>&1 || { echo "Please install Java" >&2; exit 1; }
 
-if [[ ! -f "${deps_jar}" ]]; then
-    echo "${deps_jar} not found. Running build '${build_deps_command}'"
-    if ! $build_deps_command ; then
+#if [[ ! -f "${deps_jar}" ]]; then
+#    echo "${deps_jar} not found. Running build '${build_deps_command}'"
+#    if ! $build_deps_command ; then
+#        exit "${E_BUILD_FAILED}"
+#    fi
+#fi
+
+if [[ ! -f "${jar}" ]]; then
+    echo "${jar} not found. Running build '${build_command}'"
+    if ! $build_command ; then
         exit "${E_BUILD_FAILED}"
     fi
-fi
-
-if ! $build_command ; then
-    exit "${E_BUILD_FAILED}"
 fi
 
 sparkSubmit() {
@@ -47,10 +50,11 @@ sparkSubmit() {
 # https://issues.apache.org/jira/browse/SPARK-16784
 # spark.executor.extraJavaOptions=-Dlog4j.configuration=log4j-spark.properties
 
-source "$current_dir/scripts/common_params.sh"
+source "${current_dir}/scripts/common_params.sh"
+#  --jars "${deps_jar}" \
+#
 
 sparkSubmit \
-  --jars "${deps_jar}" \
   --class "${app_class}" \
   --master "${MASTER:=local[*]}" \
   --name "${app_name}" \
@@ -58,5 +62,5 @@ sparkSubmit \
   --conf "spark.default.parallelism=8" \
   --driver-java-options "-Dlog4j.configuration=jar:file:${jar}!/log4j.properties" \
   "${jar}" \
-  "${COMMAND_PARAMS[@]}"
+  "$@"
 
--- a/query
+++ b/query
@@ -1,11 +1,10 @@
 #!/bin/bash
-# Convenience script to query for a similar files for a single file
+# Convenience script to query for a similar files of a given file
 
 E_BUILD_FAILED=142
 
-version=$(grep "version :=" build.sbt | cut -d'=' -f 2 | tr -d ' ' | tr -d '"')
-jar="target/gemini-${version}.jar"
-build_command="./sbt package"
+jar="target/gemini-uber.jar"
+build_command="./sbt assembly"
 current_dir="$(dirname "$0")"
 
 app_class="tech.sourced.gemini.cmd.QueryApp"
@@ -19,7 +18,6 @@ if [[ ! -f "${jar}" ]]; then
     fi
 fi
 
-source "$current_dir/scripts/common_params.sh"
+source "${current_dir}/scripts/common_params.sh"
 
-#exec java -cp "${jar}" "${app_class}" $@
-exec ./sbt --warn "run-main ${app_class} ${COMMAND_PARAMS[*]}"
+exec java -cp "${jar}" "${app_class}" $@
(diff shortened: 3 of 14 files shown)


bzz:
Only integration test profile on CI fails, due to networking issue on Travis side:


```
E: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/pool/main/s/snappy/libsnappy-dev_1.1.0-1ubuntu1_amd64.deb  Unable to connect to mirror.jmu.edu:http:
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?

The command "./scripts/install_python.sh" failed and exited with 127 during .
```

.gitignore
```diff
@@ -20,3 +20,5 @@ scalapbc-0.7.1.zip
 
 docfreq.json
 params.json
+
+src/test/resources/weighted-minhash/csv/
```
smacker: new line

hash
```diff
@@ -46,16 +51,16 @@ sparkSubmit() {
...
   --class "${app_class}" \
   --master "${MASTER:=local[*]}" \
   --name "${app_name}" \
+  --jars "${deps_jar}" \
   --conf "spark.executor.memory=4g" \
   --conf "spark.default.parallelism=8" \
   --driver-java-options "-Dlog4j.configuration=jar:file:${jar}!/log4j.properties" \
   "${jar}" \
-  "${COMMAND_PARAMS[@]}"
+  "$@"
```
smacker: we need COMMAND_PARAMS to read from env variables

query
```diff
@@ -1,25 +1,34 @@
...
         exit "${E_BUILD_FAILED}"
     fi
 fi
 
-source "$current_dir/scripts/common_params.sh"
+source "${current_dir}/scripts/common_params.sh"
 
-#exec java -cp "${jar}" "${app_class}" $@
-exec ./sbt --warn "run-main ${app_class} ${COMMAND_PARAMS[*]}"
+exec java -cp "${jar}:${deps_jar}" "${app_class}" $@
```
smacker: I would like to still be able to run `query` using sbt for development. It would be nice to read ENV variable and depend on it run `sbt` or `java`.
Otherwise currently for any change of source code developers must run 2-3 commands which work much slower than `./sbt package`.
What do you think?

report
```diff
@@ -1,27 +1,35 @@
...
         exit "${E_BUILD_FAILED}"
     fi
 fi
 
-source "$current_dir/scripts/common_params.sh"
+source "${current_dir}/scripts/common_params.sh"
 
-#exec java -cp "${jar}" "${app_class}" $@
-exec ./sbt --warn "run-main ${app_class} ${COMMAND_PARAMS[*]}"
+exec java -cp "${jar}:${deps_jar}" "${app_class}" $@
```
smacker: same comment as for query

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -76,7 +77,7 @@ abstract class ConnectedComponents(log: Slf4jLogger) {
         }
 
         val bucketsSize = buckets.size - prevBucketsSize
-        log.info(s"Fetched $hashtable, $bucketsSize buckets")
+        log.debug(s"Fetched $hashtable, $bucketsSize buckets")
```
smacker: let's move changes of logging to separate PR. It has nothing in common with packaging.

src/main/scala/tech/sourced/gemini/cmd/HashSparkApp.scala
```diff
@@ -55,7 +56,6 @@ object HashSparkApp extends App with Logging {
     opt[Int]("features-extractor-port")
       .action((x, c) => c.copy(fePort = x))
       .text("port is features-extractor server port")
-    opt[Unit]('v', "verbose")
```
smacker: why did you remove reading verbose flag? And I see below you still use it.

hash
bzz: Could you elaborate on it plz? Not sure I understood what you mean. Some links to code/docs would be appreciated

src/main/scala/tech/sourced/gemini/cmd/HashSparkApp.scala
bzz: It was removed in aea49edbbfe980aee8d41d326d209ff131aec9df 

If defined twice as before, it would not be updated in Configuration and will aways result in `False` and thus verbose mode was never called. This tricky behavior costed few hours of debugging.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
bzz: I prefer keeping it here as it allows to clearly see the situation before/after the packaging.

All the rest improvements except for logging are kept for the separate PR.

hash
smacker: This script includes https://github.com/src-d/gemini/blob/master/scripts/common_params.sh
That one reads env variables and sets flags.

query
bzz: I did not understood your suggestion, but from what I see right now - it's only `./sbt assembly` that is needed to be re-run on any change in code. 

`./sbt assemblyPackageDependency` would only be needed if new dependency is added.

Such developer experience looks good to me, that being said, I'm open to suggestions. 
Please, feel free to propose better developer experience in a new issue with desired behaviour and we can work from there.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: sorry, which situation? Please help me understand how internal logging in the middle of the program allows to see results of packaging?

src/main/scala/tech/sourced/gemini/cmd/HashSparkApp.scala
smacker: Wow! Thanks for the explanation! Good catch!

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
bzz: Sure - it was added along with missing logging at the beginning of the program, so you could clearly see sbt/JVM start overhead. The rest still holds.

report
bzz: although having this in .sh is VERY unnoying, logged, not to forget to get rid of this under https://github.com/src-d/gemini/issues/120#issuecomment-395741460

src/main/scala/tech/sourced/gemini/cmd/HashSparkApp.scala
bzz: Np! Was quite confused with this behaviour as well.

query
smacker: before `./query -params ...` would recompile and run the query command. It was very handy and for me, as a developer, it's really important to keep some way to run&execute.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
smacker: I do understand the need for the logging at the beginning of the program and it's already in master: https://github.com/src-d/gemini/blob/master/src/main/scala/tech/sourced/gemini/QueryApp.scala#L65

If you feel so strongly about keeping it here, please let me re-review the code to check logging too.

bzz:
CI is ✅  and feedback was addressed in 3ce9708 & 3ce9708

smacker:
@bzz github has collapsed the comment: https://github.com/src-d/gemini/pull/125#discussion_r194044389 it's really important for me and I believe for any other developer.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
bzz: This was about having it consistently in all commands. 

Please review logging as well, it's conveniently structured in a separate commits that should simplify the process.

query
bzz: As discussed offline, the suggestion was to keep best of both worlds:
 - for development of non-Spark tools: keep single command that builds and runs (`./sbt run-main`) 
 - otherwise, use `build & java -cp` way of running, same for all 3 tools

bzz: @smacker Would something like 50b52a9 make sense to you? Addressed by adding "DEV" env var.

Personally, although I appreciate interpreted languages workflow, I do not like this: those are two very different ways to organize classpath, and thus it opens the possibility of "everything works on my machine" for developer, who used `./sbt`.

But CI works on final .jars so this hopefully will be catched early.

README.md
```diff
@@ -128,34 +128,42 @@ Hash command specific arguments:
...
+This is a convenient for local development, as not requiring a separate "compile" step allows for a dev workflow
+that is similar to experience with interpreted languages.
+
+### Build
+To build final .jars for all commands
 ```
+./sbt assemblyPackageDependency
 ./sbt assembly
 ```
+Instead of 1 fatJar we bulid 2, separating all the dependencies from actuall application code to allow for
```
smacker: actuall > actual

README.md
```diff
@@ -128,34 +128,42 @@ Hash command specific arguments:
...
+lower build times in case simple changes.
```
smacker: `in case simple changes` sounds weird. Is the `of` missed?

src/main/scala/tech/sourced/gemini/cmd/QueryApp.scala
```diff
@@ -62,17 +63,20 @@ object QueryApp extends App {
...
       val cluster = Cluster.builder()
         .addContactPoint(config.host)
         .withPort(config.port)
         .build()
       val cassandra = cluster.connect()
       val gemini = Gemini(null, log, config.keyspace)
+      log.info("Checking schema")
       gemini.applySchema(cassandra)
 
+      log.info("Setting up bblfsh/fe gRCP clients")
```
smacker: rpc

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -139,6 +151,7 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
     *         - Map of element to ID
     */
   def findConnectedComponents(conn: Session): (Map[Int, Set[Int]], Map[Int, List[Int]], Map[String, Int]) = {
+    log.info("Finding Connected Componentss")
```
smacker: components

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -127,7 +136,10 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
   def reportCommunities(conn: Session,
                         communities: List[(Int, List[Int])],
                         elementIds: Map[String, Int]): Iterable[Iterable[RepoFile]] = {
-    getCommunities(conn, keyspace, communities, elementIds)
+    log.info(s"Report similar items form DB $keyspace")
```
smacker: from

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -94,7 +100,10 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
     * @return
     */
   def report(conn: Session): Iterable[Iterable[RepoFile]] = {
-    findAllDuplicateItems(conn, keyspace)
+    log.info(s"Report duplicate items form DB $keyspace")
```
smacker: from

smacker:
Please fix multiple typos in logging.

bzz:
Feedback addressed in 6ccd9c2
Merging if there is no further discussion.

PR #134, opened 2018-06-19
Refactor opts parsing

smacker:
Part of #120 about reading env vars.
The refactoring of configs isn't as easy as I would like it to be. I'll do it in separate PR.


smacker:
CI is failing because of:
```
$ go run ./src/main/go/query.go ./src/test/resources/LICENSE
# github.com/gocql/gocql
../../../gopath/src/github.com/gocql/gocql/connectionpool.go:62: sslOpts.Config.Clone undefined (type *tls.Config has no field or method Clone, but does have tls.clone)
```

it's unrelated to the PR.

bzz:
CI failure seems to be related to https://github.com/gocql/gocql/pull/1143 - thank you for the notice, will take a look.

src/main/scala/tech/sourced/gemini/cmd/Parser.scala
```diff
@@ -0,0 +1,43 @@
...
+    val cliNames = args
+      .filter(arg => arg.startsWith("-")) // remove positional arguments
+      .map(arg => arg.split("=", 2)(0))
+      .map(name => shortToFullMapping.getOrElse(name, name)) // convert short names to full names
+
+    val envNames = options
+      .map(opt => opt.fullName)
+      .filter(name => name.startsWith("--")) // remove positional arguments
+      .filterNot(name => name == "verbose") // we don't support verbose as env var
+      .filterNot(name => cliNames.contains(name)) // don't read env vars for passed flags
```
bzz: Could you please help me understand high-level logic of ENV VAR vs CLI flags?

Usual expectations from configuration subsystem are:
 - all configuration options (except input path, which is always passed though positional arg) can be set either though `ENV_VAR` or though CLI flags (`--env-var`, `-e`)
 - one form has precedence over another i.e CLI flags always override ENV VARs

This has benefits of simplicity and predictability for the end user.

Would that be the case right now or do you think there is a better option? Thanks!

smacker: yes. It works exactly as you described. I hoped comments in the code make it clear.

bzz: Nice! Thank you, was confused with that some env vars are not read.

> we don't support verbose as env var

What is the reason for that?

smacker: The code doesn't read env vars which were provided by flags already. (they will override env vars anyway but it would make code more complicated)

> we don't support verbose as env var. What is the reason for that?
currently, in master, we don't read verbose env variable. The PR is refactoring. Refactoring usually shouldn't change the way how the code works only internals.

though I don't see any problem with enabling it.

bzz: Awesome, yes, thank you again for explanation!

bzz:
Looks good, let's make sure all flags (except for positional) are accepted as env vars and :shipit: !

smacker:
I removed filter of verbose flag. Merging.

PR #140, opened 2018-06-21
Hash: apply performant defaults for PGA

bzz:
- smaller size of input data per thread
 - fix env var name in docker compose
 - add more memory for Spark in local mode (docker compose)
 - switch to efficient serialization library
 - change caching strategy (spill on disk serialized objects)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,9 +6,11 @@ services:
       - ./repositories:/repositories
       - ./query:/query
     environment:
-      DB_HOST: scylla
+      HOST: scylla
       BBLFSH_HOST: bblfshd
       FEATURES_EXTRACTOR_HOST: featurext
+    ports:
+      - "4040:4040"
 
   scylla:
     image: scylladb/scylla:2.0.0
@@ -23,6 +25,8 @@ services:
       - "127.0.0.1"
       - "--memory"
       - "2G"
+    ports:
+      - "9042:9042"
 
   featurext:
     build:
--- a/hash
+++ b/hash
@@ -46,6 +46,7 @@ sparkSubmit() {
 #  --conf "spark.local.dir=/spark-temp-data" \
 #  --conf "spark.executor.extraJavaOptions=-Djava.io.tmpdir=/spark-temp-data" \
 #  --conf "spark.eventLog.enabled=true" \
+#  --conf "spark.eventLog.dir=/repositories"
 
 # https://issues.apache.org/jira/browse/SPARK-16784
 # spark.executor.extraJavaOptions=-Dlog4j.configuration=log4j-spark.properties
@@ -55,6 +56,7 @@ sparkSubmit \
   --master "${MASTER:=local[*]}" \
   --name "${app_name}" \
   --jars "${deps_jar}" \
+  --conf "spark.driver.memory=4g" \
   --conf "spark.executor.memory=4g" \
   --conf "spark.default.parallelism=8" \
   --conf "spark.tech.sourced.engine.skip.read.errors=true" \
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -6,9 +6,11 @@ services:
       - ./repositories:/repositories
       - ./query:/query
     environment:
-      DB_HOST: scylla
+      HOST: scylla
       BBLFSH_HOST: bblfshd
       FEATURES_EXTRACTOR_HOST: featurext
+    ports:
```
smacker: one of the reasons why there are no ports - make sure user can run gemini without port collisions.

hash
```diff
@@ -55,8 +56,12 @@ sparkSubmit \
   --master "${MASTER:=local[*]}" \
   --name "${app_name}" \
   --jars "${deps_jar}" \
+  --conf "spark.driver.memory=4g" \
```
smacker: does it mean we require 4GB of memory to be given to docker?
If yes we need at least to document it. For "docker for mac" the default value is 1GB if I remember right.

src/main/scala/tech/sourced/gemini/Hash.scala
```diff
@@ -33,8 +34,8 @@ class Hash(session: SparkSession, log: Slf4jLogger) {
     * @param repos DataFrame with engine.getRepositories schema
     */
   def forRepos(repos: DataFrame): HashResult = {
-    val files = filesForRepos(repos).cache()
-    val uasts = extractUast(files).cache()
+    val files = filesForRepos(repos).persist(StorageLevel.MEMORY_AND_DISK_SER)
+    val uasts = extractUast(files).persist(StorageLevel.MEMORY_AND_DISK_SER)
```
smacker: should we make this option configurable?

hash
```diff
@@ -55,8 +56,12 @@ sparkSubmit \
...
   --conf "spark.executor.memory=4g" \
   --conf "spark.default.parallelism=8" \
+  --conf "spark.serializer=org.apache.spark.serializer.KryoSerializer" \
```
smacker: According to the [docs](https://spark.apache.org/docs/latest/tuning.html#data-serialization) KryoSerializer requires classes registration. Should we at least create an issue for that?

docker-compose.yml
bzz: I see. But without having this we would not be able to point users of `docker-compose` to Spark web ui, which is crucial for debugging.

hash
bzz: No, it does not. I just max capacity that the Driver JVM process possibly can take.

src/main/scala/tech/sourced/gemini/Hash.scala
bzz: We will, one day, after fine-tuning and understanding performance implications better. For now, it just changes the default persistence level from one to another.

hash
bzz: 👍 this is not mandatory but indeed will save some space later on https://github.com/src-d/gemini/issues/141

bzz:
> the main concern is how gemini will work with default configuration on docker

Valid concern, thank you for raising it. In short - it will not. 

Long story: you can see it yourself - a [quick search](https://github.com/apache/spark/search?q=spark+driver+memory&unscoped_q=spark+driver+memory) over Spark source code reveals that [this option](https://github.com/apache/spark/blob/f5fe63f7b8546b0102d7bfaf3dde77379f58a4d1/launcher/src/main/java/org/apache/spark/launcher/SparkClassCommandBuilder.java#L111) will set JVM `-Xmx` argument that is a standard for setting the [maximum heap size](https://www.eclipse.org/openj9/docs/xms/) of the process.

Hope this addresses the concern.

docker-compose.yml
smacker: I would suggest such users use `.host.yaml` config. But okay. The only problem I'll have to edit `docker-compose.yml` every time when I want to test it with docker-compose but I can live with it.

PR #142, opened 2018-06-24
Use JSON parser

bzz:
`scala.util.parsing.json` is deprecated and known to be not production ready parser implementation.

This PR replaces it with Jackson, that has performance benefit at Apache Spark case - it avoid copying 100s Mb object on the Driver.


bzz:
> not sure why terminate gRPC connection is here 

it's just because it has been noticed while working on this, and is small enough that does not worth overhead of having another PR

PR #145, opened 2018-06-25
Hash: refactoring of feature hashing

bzz:
Based on #142  (only last 3 commits are new)

 - [x] This changes client API to make explicit dependency on expensive WMH data-structure,
instead of a keeping it in a global static state.

   This allows for more fine-grain control over resources and speeds up hashing ~2.5 times by removing excessive synchronization between task threads in executors processes on Spark.

 - [x] Broadcast DocFreq to avoid sending it \w every task (~100Mb for 1 PGA bucket)

### DIFF
--- a/src/main/scala/tech/sourced/gemini/FeaturesHash.scala
+++ b/src/main/scala/tech/sourced/gemini/FeaturesHash.scala
@@ -2,7 +2,6 @@ package tech.sourced.gemini
 
 import tech.sourced.featurext.generated.service.Feature
 
-import scala.collection.mutable
 import scala.collection.Searching._
 
 object FeaturesHash {
@@ -44,10 +43,10 @@ object FeaturesHash {
     * @param docFreq
     * @return
     */
-  private def toBagOfFeatures(features: Iterable[Feature], docFreq: OrderedDocFreq): Array[Double] = {
+  def toBagOfFeatures(features: Iterable[Feature], docFreq: OrderedDocFreq): Array[Double] = {
     val OrderedDocFreq(docs, tokens, df) = docFreq
 
-    val bag = new Array[Double](tokens.size)//mutable.ArrayBuffer.fill(tokens.size)(0.toDouble)
+    val bag = new Array[Double](tokens.size) //mutable.ArrayBuffer.fill(tokens.size)(0.toDouble)
     features.foreach { feature =>
       tokens.search(feature.name) match {
         case Found(idx) => {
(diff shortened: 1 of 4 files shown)


src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -44,10 +43,10 @@ object FeaturesHash {
     * @param docFreq
     * @return
     */
-  private def toBagOfFeatures(features: Iterable[Feature], docFreq: OrderedDocFreq): Array[Double] = {
+  def toBagOfFeatures(features: Iterable[Feature], docFreq: OrderedDocFreq): Array[Double] = {
     val OrderedDocFreq(docs, tokens, df) = docFreq
 
-    val bag = new Array[Double](tokens.size)//mutable.ArrayBuffer.fill(tokens.size)(0.toDouble)
+    val bag = new Array[Double](tokens.size) //mutable.ArrayBuffer.fill(tokens.size)(0.toDouble)
```
smacker: not sure what this comment is doing here. Most probably some leftover...

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
```diff
@@ -13,22 +13,20 @@ object FeaturesHash {
...
-      tokensSize = docFreq.tokens.size
-    }
-
-    val bag = toBagOfFeatures(features, docFreq)
-    wmh.hash(bag)
+  /**
+    * Factory method for initializing WMH data structure
+    * \w default parameters, specific to Gemini.
+    *
+    * Allocates at lest 2*dim*sampleSize memory
```
smacker: `at least`

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -139,8 +139,12 @@ class FileQuery(conn: Session,
     log.info(s"Reading docFreq from ${docFreqFile.getAbsolutePath}")
     val docFreq = OrderedDocFreq.fromJson(docFreqFile)
 
+    log.info(s"Initialize WMH for ")
```
smacker: Initialize WMH for **what**?

src/main/scala/tech/sourced/gemini/FeaturesHash.scala
bzz: Yup, seems like a leftover. Will remove

bzz:
Rebased and fixed, waiting for CI to be green before merging.

Issue #112, opened 2018-05-04
Save docFreq and params inside the DB


bzz:
Propose a DB schema for saving docFreq and params inside DB, instead of .json in FS.

Then migrate:
   * hash, to writing to DB 
   * query, to reading params from DB

bzz:
One thing I'd like to understand better is: given the same vocabulary (and thus re-using a docFreq) it is reasonable to expect a user to re-run `./hash` multiple times with different similarity threshold settings, thus, generating new params. And then later using one of those for `./report` or `./query`.

Would this use-case change anything in DB vs Files scenario? Pointing to a separate files through CLI is very intuitive.

At the same time, hash would produce a different DBs with hashes on every run, so if we do accommodate such use-case only by switching a full `--keyspaces` now - it does not matter that much.

smacker:
It's a bad idea to use different docFreq with the same DB (with multiple re-run of hash). It might start to produce wrong results because the same files will have multiple hashes in DB (for each new docFreq). Then query and report would use hashes from old runs too. (not sure how real collisions are in real life, but still).
Switching namespaces is better choice for such case.

bzz:
Agreed.

This would also allow us to get rid of using [scala.util.parsing.json](https://github.com/scala/scala-parser-combinators/blob/8e8e88042c20a327464c71b1376de72c72ec0bd7/shared/src/main/scala/scala/util/parsing/json/package.scala#L4) which sounds very good due to:
   
>  This package was never intended for production use; it's really more of a code sample demonstrating how to use parser combinators.

smacker:
Schema:

```
CREATE TABLE docfreq (
    id text PRIMARY KEY,
    docs int,
    df map<text, int>
);
```

map is a sorted type already according to the docs: http://docs.scylladb.com/getting-started/types/#maps

for now, we can hard-code `id` but it can be useful to be able to save more than 1 docfreq in the future.

PR #151, opened 2018-07-02
Docfreq to db

smacker:
Fix: #112

The schema was described in the issue.
I kept reading docFreq from JSON for test.

### DIFF
--- a/src/main/scala/tech/sourced/gemini/FileQuery.scala
+++ b/src/main/scala/tech/sourced/gemini/FileQuery.scala
@@ -13,7 +13,6 @@ import tech.sourced.featurext.generated.service.FeatureExtractorGrpc.FeatureExtr
 import tech.sourced.gemini.util.MathUtil
 
 import scala.collection.JavaConverters._
-import scala.collection.JavaConversions.mapAsScalaMap
 import scala.util.control.NonFatal
 
 /**
@@ -137,8 +136,10 @@ class FileQuery(conn: Session,
     if (row == null) {
       None
     } else {
-      val javaDf = row.getMap("df", classOf[java.lang.String], classOf[java.lang.Integer])
-      val df = javaDf.mapValues(Integer2int(_))
+      val df = row
+        .getMap("df", classOf[java.lang.String], classOf[java.lang.Integer])
+        .asScala
+        .mapValues(_.toInt)
 
       Some(OrderedDocFreq(row.getInt(cols.docs), df.keys.toIndexedSeq, df))
     }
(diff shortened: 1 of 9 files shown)


bzz:
Nice job! If you do not mind, let me review this change a bit later this week.

bzz:
As discussed offline, I suggest keeping the `--doc-freq-file` CLI flag, so it triggers saving/reading from JSON and switching to DB for default case when no flag was provided.

This simplifies debugging and comparing experiments results with the previous runs and also make possible fast migration with another DB, with no Map datatype support.

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -13,6 +13,7 @@ import tech.sourced.featurext.generated.service.FeatureExtractorGrpc.FeatureExtr
 import tech.sourced.gemini.util.MathUtil
 
 import scala.collection.JavaConverters._
+import scala.collection.JavaConversions.mapAsScalaMap
```
bzz: It's better to avoid using this old conversions API and stick to a new one from `JavaConverters` imported above instead \w it's `.asScala` .

src/main/scala/tech/sourced/gemini/FileQuery.scala
```diff
@@ -136,10 +130,21 @@ class FileQuery(conn: Session,
...
-    val docFreq = OrderedDocFreq.fromJson(docFreqFile)
+  protected def readDocFreq(): Option[OrderedDocFreq] = {
+    log.info(s"Reading docFreq from DB")
+    val cols = tables.docFreqCols
+    val row = conn.execute(s"SELECT * FROM ${tables.docFreq} WHERE ${cols.id} = '1'").one()
+    if (row == null) {
+      None
+    } else {
+      val javaDf = row.getMap("df", classOf[java.lang.String], classOf[java.lang.Integer])
+      val df = javaDf.mapValues(Integer2int(_))
```
bzz: Related to [comment on Java/Scala converters](https://github.com/src-d/gemini/pull/151/files#r200662238) - without changing the way DB is queried, it would be more natural to have something like

```scala
      val df = row
        .getMap("df", classOf[java.lang.String], classOf[java.lang.Integer])
        .asScala
        .mapValues(_.toInt)
```

src/main/scala/tech/sourced/gemini/Hash.scala
```diff
@@ -1,5 +1,8 @@
 package tech.sourced.gemini
 
+
+import com.datastax.spark.connector.cql.CassandraConnector
+import collection.JavaConversions._
```
bzz: Same as above, lets sick with ` import scala.collection.JavaConverters._`

smacker:
CI is red because of service 'featurext' failed to build. Not related to PR.

smacker:
rebased on fixed master

smacker:
@bzz Are we okay with merging it? I see a thumb up on the last comment but no direct approval.

bzz:
As there is no time pressure, let me get back to you on this a bit later.

Issue #129, opened 2018-06-13
Refactoring: get list of supported languages from bblfsh


bzz:
Now, that https://github.com/bblfsh/bblfshd/releases/tag/v2.5.0 is released, let's
 - after https://github.com/bblfsh/client-scala/issues/68 is resolved
 - do https://github.com/src-d/gemini/blob/632680701f95cf9f5620dd86bf61fcca99751b0e/src/main/scala/tech/sourced/gemini/Hash.scala#L66
 - update https://github.com/src-d/gemini/blob/632680701f95cf9f5620dd86bf61fcca99751b0e/docker-compose.yml#L35
 - update README, documenting min bblfshd version compatibility

smacker:
bblfsh/client-scala#68 was resolved https://github.com/bblfsh/client-scala/pull/69

smacker:
Do we really need to filter it on our side?

There is a PR to engine https://github.com/src-d/engine/pull/404 which would return empty UAST to unsupported languages so they will be filtered by current
```
.filter(_.getAs[Seq[Array[Byte]]]("uast").nonEmpty)
```

bzz:
Yup, we can skip filtering on our side now

PR #153, opened 2018-07-10
bump engine to 0.7.0

smacker:
Related things for us:
- support for Apache Spark 2.3 https://github.com/src-d/engine/pull/403
- bblfsh: do not send requests for unsupported languages https://github.com/src-d/engine/pull/404

Looks like the last one allows to remove list of languages from gemini completely https://github.com/src-d/gemini/issues/129

PR #154, opened 2018-07-10
Remove language filtering

smacker:
Based on #153 
Fix: #129


bzz:
👍 Let's also take care of updating a deployment with appropriate bblfshd version, before merging this.

Otherwise we would not be able to run master on staging cluster with https://github.com/src-d/issues-infrastructure/issues/184

smacker:
@bzz We aren't able to run to run master on staging cluster already because the new engine is in master now. And it isn't compatible with old bblfsh. This PR doesn't change it.

bzz:
Yep, please, let's fix it and link an Infra issue in there and #154  for convenience.

smacker:
Yes. The issue is here: https://github.com/src-d/issues-infrastructure/issues/201

But I don't see any reason to wait for it to merge this PR. Master is broken already.

PR #157, opened 2018-10-01
Refactoring/extractor params

bzz:
Based on #155 (only fc20930 is new)

Parametrize set configuration for active set of Feature Extractors

```scala
  val fileLevelExtractors = Seq(IdentifiersExt(194, split = true), GraphletExt(548), LiteralsExt(264))
```

Issue #150, opened 2018-07-02
Function-similarity: apply default parameters, found by ML team


EgorBu:
Parameters for val_acc 91.6% (used for optimization) and test_acc 90.6% (hold out):
* best combination of feature extractors`('id', 'uast2seq', 'graphlet')` (it looks like that structural information even more important for functions because of lack of literals)
* `id_split_stem True` - identifiers should be splitted
* `'uast2seq_weight': 369`
* `''id_weight': 535`
* `graphlet_weight': 5707`
* `'threshold': 0.592`
* `hash size 256`
* `'uast2seq_seq_len': 3`
* `'uast2seq_stride': 1`

bzz:
@EgorBu thanks for detailed info!

Feature extractor parameters were taken care of  https://github.com/src-d/gemini/pull/160 the rest is tracked under #162

smacker:
fixed

PR #160, opened 2018-10-03
Function-level similarity: CLI mode + function extraction

bzz:
Based on #155 and #157 

This is implemented part of the #150

TODO:
 - [x] CLI `--mode func|file` for function/file level ops
 - [x] hash: function extraction from UAST
 - [x] hash: default FE parameters for function/file level ops from ML team

Further improvements on hashing and query sides are tracked under #162


src/main/scala/tech/sourced/gemini/cmd/HashSparkApp.scala
```diff
@@ -23,14 +23,15 @@ case class HashAppConfig(
...
+  verbose: Boolean = false,
+  mode: String = Gemini.fileSimilarityMode
 )
 
 /**
   * Apache Spark app that applied LSH to given repos, using source{d} Engine.
   */
 object HashSparkApp extends App with Logging {
-  val repoFormats = Seq("siva", "bare", "standard")
+  val repoFormats = Seq(Gemini.fileSimilarityMode,  Gemini.funcSimilarityMode)
```
carlosms: Bug, right? `repoFormats` should not change, it's still used for `--format`, format of the stored repositories.

carlosms:
I would maybe remove the `list_funcs.py`. If it's only for reference, having it in `src/main/python` with other code we do use will be confusing in the future.

Plus `extractFunctions` already has a link to the exact revision used as reference.

src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -27,7 +27,7 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
     * @param limit
     * @param format
     */
-  def hash(reposPath: String, limit: Int = 0, format: String = "siva", docFreqPath: String = ""): Unit = {
+  def hash(reposPath: String, limit: Int = 0, format: String = "siva", mode: String, docFreqPath: String = ""): Unit = {
```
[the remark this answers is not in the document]
bzz: added in 89f2028

bzz:
@carlosms all feedback addressed, PR description update with current scope, ready for another pass.

src/test/scala/tech/sourced/gemini/SparkHashSpec.scala
```diff
@@ -36,11 +36,28 @@ class SparkHashSpec extends FlatSpec
...
+  "Hash" should "return correct number of functions" in {
+    val hashResult = hashWithNewGemini("func")
+    val funcs = hashResult.hashes
+    val fCount = funcs.count()
+    print(s"$fCount functions found")
+    funcs.select("doc").show(false)
+    fCount shouldEqual 54
+  }
+
+  "Hash" should "extract functions UASTs in func mode" in {
```
smacker: this test doesn't have any assert

Issue #159, opened 2018-10-03
2 tests failing on current master


bzz:
As of d111064b4320ca9860f4f3f6829688a8026baf64 2 tests are failing on master (test on 2 commits before it, fails the same way).

Moved from https://github.com/src-d/gemini/pull/158#issuecomment-426233471

## I. hash

```
 WARN 00:06:35,582 tech.sourced.gemini.Hash (Hash.scala:36) - Feature Extraction exceptions
 WARN 00:06:35,582 tech.sourced.gemini.Hash (Hash.scala:39) - Processed: 1671, skipped: 0
 WARN 00:06:35,582 tech.sourced.gemini.Hash (Hash.scala:123) - creating document frequencies
 WARN 00:06:38,188 tech.sourced.gemini.Hash (Hash.scala:139) - hashing features
[info] TEST FAILED - SparkHashSpec: Hash should generate docFreq: 773 did not equal 738
[info]   773 did not equal 738 (SparkHashSpec.scala:73)
```
was passing last time [3 months ago](https://travis-ci.org/src-d/gemini/jobs/402705581) on [Jul 11](https://github.com/src-d/gemini/commit/2c195660de594ba6e22ab37d48859cb66ac5bed5)

```
 WARN 15:05:34,072 tech.sourced.gemini.Hash (Hash.scala:34) - Feature Extraction exceptions
 WARN 15:05:34,072 tech.sourced.gemini.Hash (Hash.scala:37) - Processed: 1550, skipped: 0
 WARN 15:05:34,073 tech.sourced.gemini.Hash (Hash.scala:116) - creating document frequencies
 WARN 15:05:36,783 tech.sourced.gemini.Hash (Hash.scala:132) - hashing features
```

fixtures under `src/test/resources/siva/duplicate-files` did not change but the number of Feature extracted is different.

The same test fails locally even on 2c195660de594ba6e22ab37d48859cb66ac5bed5 where CI did pass 3 months ago, even with `bblfshd/go-driver:v0.4.0` (that was `:lastest` back then)

## Community detection

Same for 
```
$ pytest -v src/main/python/community-detector

src/main/python/community-detector/test_community_detector.py:46: AssertionError

       assert_array_equal(data, fixture_data)
E       AssertionError:
E       Arrays are not equal
E
E       (mismatch 6.201550387596896%)
```

smacker:
Last master is green: https://travis-ci.org/src-d/gemini/builds/439166300

PR #163, opened 2018-10-08
Update fixutres to fix python report tests

smacker:
Part of #159

The main issue was that fixtures were created with python3.4 but now we
use python3.6. It cases small mismatch (6.201550387596896%) in results.

This commit update fixtures and sets exact versions for sourced-ml and
modelforge packages.

Signed-off-by: Maxim Sukharev <maxim@sourced.tech>


smacker:
CI for python report is green now.

Issue #162, opened 2018-10-05
Function-level similarity


bzz:
This is second part of the #150 

Initial work on function extraction and introducing `--mode file|func` was done in #160

https://github.com/src-d/gemini/pull/160

TODOs:
 - [x] FE: add Graphlet args (seqLen,stride) to .proto, bump grpc to 0.13, re-generate all protos and fixtures https://github.com/src-d/gemini/pull/166
 - [x] hash: calculate `HashtablesNum`/`BandSize` for `sample_size=256` and `threshold=0.592` using a [helper script](https://github.com/src-d/gemini/pull/116#issuecomment-387386955) #168
 - [x] hash: parametrize FeaturesHash.scala \w those 2 params, similar to what was done to FE in 1a0d00b #168
 - [x] hash: save results to DB. Either replacing files, or to another table. #172 #174
 - [x] query: choose right configuration for FE and hashing \w params, depending on `--mode` #169
 - [ ] query: update CLI if needed, e.g. `./query ./filename_funcName:lineNumber`
 - [x] report: either include both file/func in same report or add `--mode` #171

PR #169, opened 2018-10-11
Similar func query

smacker:
Part of https://github.com/src-d/gemini/issues/162

sorry for reformat of yaml file. It happened automatically. (though I think this format is easier to understand)


carlosms:
LGTM, but it would need some docs for the new functionality.

carlosms:
Makes sense :+1:

PR #171, opened 2018-10-16
report implementation for similar funcs

smacker:
Part of #162

TODO:
- integration test
- unit test
- check that it works if hash command was called for both funcs and files


smacker:
rebased on master

PR #172, opened 2018-10-17
Docfreq type

smacker:
Part of #162 (check the first comment)


src/main/scala/tech/sourced/gemini/Hash.scala
```diff
@@ -34,9 +34,13 @@ case class HashResult(files: DataFrame, hashes: Dataset[RDDHash], docFreq: Order
   *
   * @param session     spark session
   * @param log         logger implementation
+  * @param mode        filse or functions hashing
```
carlosms: Typo, "files".

Also, should we mention it must be one of `Gemini.fileSimilarityMode`, `Gemini.funcSimilarityMode`?

PR #176, opened 2018-10-22
Pass func name and line to query

### LINKED ISSUE #162: Function-level similarity

smacker:
Fix #162

The last item in the list. Adds support for querying only 1 function in a file:
```
/query ./filename.ext:funcName:lineNumber
```

### DIFF
--- a/src/main/scala/tech/sourced/gemini/Gemini.scala
+++ b/src/main/scala/tech/sourced/gemini/Gemini.scala
@@ -89,11 +89,12 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
             mode: String,
             docFreqPath: String = "",
             feClient: FeatureExtractor): QueryResult = {
+    log.info(s"Query for items similar to $inPath")
     val (path, fnFilter) = fileWithFuncPattern.findFirstMatchIn(inPath) match {
       case Some(m) => (new File(m.group(1)), Some(m.group(2), m.group(3).toInt))
       case None => (new File(inPath), None)
     }
-    log.info(s"Query for items similar to $path")
+
     if (path.isDirectory) {
       QueryResult(findDuplicateProjects(path, conn, keyspace), findSimilarProjects(path))
     } else {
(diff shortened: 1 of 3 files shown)


src/main/scala/tech/sourced/gemini/Gemini.scala
```diff
@@ -86,13 +89,16 @@ class Gemini(session: SparkSession, log: Slf4jLogger, keyspace: String = Gemini.
             mode: String,
             docFreqPath: String = "",
             feClient: FeatureExtractor): QueryResult = {
-    val path = new File(inPath)
+    val (path, fnFilter) = fileWithFuncPattern.findFirstMatchIn(inPath) match {
+      case Some(m) => (new File(m.group(1)), Some(m.group(2), m.group(3).toInt))
+      case None => (new File(inPath), None)
+    }
     log.info(s"Query for items similar to $path")
```
carlosms: I guess we should also log `fnFilter`?

smacker: fixed

smacker:
Updated.

PR #183, opened 2019-01-24
add support for Amazon Web Services S3

smacker:
As usual, it took a while to figure which versions should be used to avoid conflicts.
s3a available since Hadoop 2.6 and considered stable since Hadoop 2.7

Allow all URI supported by hadoop-aws:
- s3:// - deprecated
- s3n:// -deprecated
- s3a:// - recommended

The URI may contain AWS key & secret (as login & password) for the
simplest case.

Similar to GCS hadoop-aws included into fat deps jar to avoid conflicts.


smacker:
*Don't merge until we test it with real S3 account.*

campoy:
What do you need to test this?
@smola should be able to test this on our PoC accounts (or maybe even provide credentials for you)

smacker:
we don't have any mention about hdfs, gcs or aws in readme right now. Let's create an issue for that?

smacker:
It's not really about "right now" but I would prefer to describe all storages in one task than doing just s3 here. I created an issue: https://github.com/src-d/gemini/issues/186

smacker:
rebased on master

Issue #186, opened 2019-01-29
Document ability to work with distrubited date sources


smacker:
Currently gemini supports:
- hdfs
- GCS
- S3

We need to document how to use each of them.

PR #187, opened 2019-01-29
add documentation for supported distributed storages

smacker:
This commit describes supported storages with instructions how to
connect to them locally and link to cluster mode configuration.

Fix: #186 

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ To pre-process repositories for search of similar functions run:
 ./hash -m func ./src/test/resources/siva
 ```
 
+Besides local file system gemini support different [distributed storages](#distributed-storages).
+
 ### Query
 
 To find all duplicate of the single file run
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,6 +161,38 @@ Report specific arguments:
...
+
+### Distributed storages
+
+Gemini supports different distributed storages in local and cluster mode. It already includes all necessary jars as a part of fat jar.
+
+#### HDFS
+
+Path format to git repositories: `hdfs://hdfs-namenode/path`
+
+To configure HDFS in local or cluster mode please consult [Hadoop documentation](https://hadoop.apache.org/docs/r2.7.0/index.html).
```
se7entyse7en: I noticed that [here](https://github.com/src-d/gemini/blob/master/project/Dependencies.scala#L10) we're using `v2.7.2` of the client, does the client follow the same versioning? In that case we should point to the corresponding doc.

PR #188, opened 2019-01-29
add Extract method to feature-extractor server

smacker:
Currently, we call feature-extractor 3 times per file in file mode and
`number of functions * number of files * 3` in function mode.

Each time we send full uast to feature-extractor server.

This commit introduces new Extract method that can run different
extractors sequentially on UAST and return features from all of them.

I kept all previous methods both in server and client but changed
request parameters so Extract and old method could share extractors
configuration.


smacker:
Please let me know if you think I should drop per-extractor methods. It would simplify the code.

src/main/proto/service.proto
```diff
@@ -17,35 +20,59 @@ service FeatureExtractor {
...
-  gopkg.in.bblfsh.sdk.v1.uast.Node uast = 1;
+message Uast2seqOptions {
   int32 docfreqThreshold = 2;
   int32 weight = 3;
   int32 stride = 4;
   repeated int32 seqLen = 5;
 }
 
+message GraphletOptions {
+  int32 docfreqThreshold = 2;
```
carlosms: I guess it's ok to skip `1` and keep the same order as when this was in `GraphletRequest`, but is there a reason to do it?

smacker: nope, no reason. I missed it because to copy-past. Thanks!

src/main/python/feature-extractor/server.py
```diff
@@ -20,53 +21,78 @@
...
+    def Extract(self, request, context):
+        """ Extract features using multiple extrators """
+
+        extractors = []
+
+        for name in self.extractors_names:
+            if request.HasField(name):
+                options = getattr(request, name, None)
+                constructor = getattr(self, "_%s_extractor" % name)
+                extractors.append(constructor(options))
```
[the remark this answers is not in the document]
smacker: as far as I understand generated grpc code it can't be None if `request.HasField` returns `true`. But I added just check just in case something changes. Thanks!

src/main/proto/service.proto
smacker: fixed

src/main/python/feature-extractor/server.py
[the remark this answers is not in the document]
se7entyse7en: Oh ok, didn't know that! But yup, better be sure by still adding the check 👍

PR #193, opened 2019-02-07
use workers pool in feature-extractor

smacker:
Feature extraction is CPU bound task.
With increase of grpc workers it hits GIL.

### DIFF
--- a/src/main/python/feature-extractor/server.py
+++ b/src/main/python/feature-extractor/server.py
@@ -74,7 +74,7 @@ class Service(service_pb2_grpc.FeatureExtractorServicer):
     }
 
     def __init__(self, pool):
-        super(Service, self)
+        super(Service, self).__init__()
         self.pool = pool
 
     def Extract(self, request, context):
--- a/src/main/python/feature-extractor/server.py
+++ b/src/main/python/feature-extractor/server.py
@@ -3,9 +3,11 @@ import sys
 sys.path.append('./pb')
 
 from concurrent import futures
+from multiprocessing import Pool
 import argparse
 import logging
 import os
+import signal
 import time
 
 import grpc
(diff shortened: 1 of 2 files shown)


smacker:
Performance testing:

Server started with 30 grpc workers.
In multiprocessing mode it had 5 workers in the pool.

3000 calls with 30 concurrent requests:
without multiprocessing: finished in 39.666251747s
with multiprocessing: finished in 23.718774715s

src/main/python/feature-extractor/server.py
```diff
@@ -14,93 +16,133 @@
...
+    pool = None
+    extractors = {
+        "identifiers": _identifiers_extractor,
+        "literals": _literals_extractor,
+        "uast2seq": _uast2seq_extractor,
+        "graphlet": _graphlet_extractor,
+    }
+
+    def __init__(self, pool):
+        super(Service, self)
```
se7entyse7en: Are you that this is working? This should be:

```suggestion
        super(Service, self).__init__()
```

Maybe it works because in the `service_pb2_grpc.FeatureExtractorServicer` class the init is actually a noop.

src/main/python/feature-extractor/server.py
```diff
@@ -110,19 +152,21 @@ def serve(port, workers):
         while True:
             time.sleep(_ONE_DAY_IN_SECONDS)
     except KeyboardInterrupt:
+        pool.terminate()
```
se7entyse7en: This won't be called for exceptions different from `KeyboardInterrupt`. Actually I don't know whether it is going to leak so resources. Maybe you could use [this](https://docs.python.org/3.7/library/atexit.html).

src/main/python/feature-extractor/server.py
smacker: 🤦‍♀️ fixed

src/main/python/feature-extractor/server.py
smacker: All sub-processes are created using fork. Operation system is going to take care of cleaning up when the server exits. So it's not a big deal. Also, we run feature-extractor in the container and any exit would cause restart of the container which would clean-up everything. 

I don't think we need to improve termination here as long as we don't have any real issue with how it works right now.

PR #194, opened 2019-02-07
create WMH only once and broadcast

smacker:
An instance of whm is relatively huge and slow to create
On ~500 middle-size repositories it consumers ~4G of memory

In my tests it causes lots of CPU consumption and currently on my dataset worker just stops responding to heartbeat checks.


se7entyse7en:
I never used broadcast, it's really interesting. 👍 LGTM.

PR #196, opened 2019-02-08
Use floats instead of doubles in wmh

smacker:
- 32 bit precision is enough for us
- we drop the rest when we calculate hashtables anyway
- change should reduce required memory for wmh instance twice


smacker:
FileQuery test has failed but with manual testing everything works file. I'm looking into it. (most probably the problem is in the test itself)

smacker:
Fixed. I had to update all related fixtures not only hashtables because go-driver started returning a little bit different results -> different features -> different docfreq.

PR #197, opened 2019-02-11
Different optimizations for report

smacker:
Please look at each commit message for details.

### DIFF
--- a/src/main/python/community-detector/community_detector.py
+++ b/src/main/python/community-detector/community_detector.py
@@ -1,4 +1,3 @@
-from collections import defaultdict
 from itertools import chain
 import logging
 
@@ -34,30 +33,7 @@ def build_matrix(id_to_buckets):
     return csr_matrix((data, indices, indptr))
 
 
-def build_id_to_cc(connected_components, length):
-    """Builds a ndarray that associates element id to a connected component
-
-    Same code as in Apollo ConnectedComponentsModel.
-    https://github.com/src-d/apollo/blob/f51c5a92c24cbedd54b9b30bab02f03e51fd27b3/apollo/graph.py#L28
-
-    Args:
-        connected_components: list of tuples (connected-component, element ids)
-        length: number of elements
-
-    Returns:
-        A 1 dimension ndarray. The index will be the element id, and the
-            value is the connected component
-    """
-
-    id_to_cc = numpy.zeros(length, dtype=numpy.uint32)
-    for cc, ids in connected_components:
-        for id_ in ids:
-            id_to_cc[id_] = cc
-
-    return id_to_cc
-
-
-def detect_communities(cc,
+def detect_communities(ccs,
                        buckets_matrix,
                        edges="linear",
                        algorithm="walktrap",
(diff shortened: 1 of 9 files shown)


smacker:
Python stylecheck is failing but it also failing on master (weird, all prs passed it. Most probably something changed in the linter). So it shouldn't block this pr.

src/main/scala/tech/sourced/gemini/ConnectedComponents.scala
```diff
@@ -52,40 +52,74 @@ abstract class ConnectedComponents(log: Slf4jLogger) {
...
+    (buckets, elementIds)
+  }
+
+  private def bucketsForHashTable(hashTable: Byte, elementIds: Map[String, Int]): List[List[Int]] = {
+    var buckets = List[List[Int]]()
+    var band: Option[ByteBuffer] = None
+    var bucket = List[Int]()
+
+    getHashValues(hashTable).foreach { case FileHash(sha1, value) =>
+      val elId = elementIds(sha1)
```
se7entyse7en: The assumption here is that it's not possible that there's an element id in the non-first hashtable that is actually not present in the first hashtable, right?

smacker: yes. WMH produces one long hash for a file. Then we split it to "bands" so each file has number of hashtables rows with partial hash in it.

smacker:
I have reworked the PR a lot. Please, another pass. (sorry Marvin)
//cc @se7entyse7en @carlosms 

P.S. On test dataset report time went down from 60+ hours (I stopped it on 3rd day of running) to 15 minutes.

se7entyse7en:
@smacker unfortunately I barely know what `gemini` does, so as in the previous reviews I just reviewed things that are narrower in scope. Maybe it worth for me taking some time to also take a deeper look at `gemini` for future reviews, but I'm afraid that if I do this now it would take much time and would delay these PRs for the POC.

src/main/python/community-detector/community_detector.py
```diff
@@ -10,11 +9,8 @@
 def build_matrix(id_to_buckets):
     """Builds a CSR matrix from a list of lists of buckets
```
carlosms: Minor, but should this be changed? Now `id_to_buckets` is a "list of tuples (element_id, buckets)" right?

src/main/python/community-detector/community_detector.py
```diff
@@ -23,41 +19,20 @@ def build_matrix(id_to_buckets):
...
+    indptr = numpy.zeros(max_el_id + 2, dtype=numpy.uint32)
     pos = 0
-    for i, element in enumerate(id_to_buckets):
-        indices[pos:(pos + len(element))] = element
-        pos += len(element)
-        indptr[i + 1] = pos
+    for el_id, bucket in id_to_buckets:
+        indices[pos:(pos + len(bucket))] = bucket
+        pos += len(bucket)
+        indptr[el_id + 1:] = pos
```
carlosms: Maybe there is a performance gain we can squeeze here, avoiding to write always up to the end of the array.
Something like this (did not test this, please double check)
```python
    prev_el_id = 0
    prev_pos = 0

    for el_id, bucket in id_to_buckets:
        indices[pos:(pos + len(bucket))] = bucket
        pos += len(bucket)
        indptr[prev_el_id+2:el_id] = prev_pos

        prev_el_id = el_id
        prev_pos = post
    
    indptr[prev_el_id+1:] = prev_pos
```

smacker: thanks! It's a good idea. I re-wrote it a little bit different and added one more test to be sure it works correctly for an edge case.

smacker:
@se7entyse7en thanks anyway for your valuable review. Carlos knows internal of gemini better and he approved the code so I think we are good to merge now without you going deep into details.

PR #224, opened 2019-06-06
A small fixup around bblfsh client

bzz:
A drive-by improvement while looking at Gemini usage of bblfsh client - use old bblfsh client version, but the one that exposes `.close()`


smacker:
thanks

